PR #3, opened 2024-11-01
v0: fname fetcher

suurkivi:
Initial version of code to fetch fnames and store them in memory (for now).

This also uses the `tracing` logger and enables json output.

When running, you can see lines like:

```
{"timestamp":"2024-11-02T01:22:35.272756Z","level":"DEBUG","fields":{"message":"fetching transfers","url":"https://fnames.farcaster.xyz/transfers?from_id=685400"},"target":"snapchain::connectors::fname"}
{"timestamp":"2024-11-02T01:22:35.751982Z","level":"INFO","fields":{"message":"found new transfers","count":100,"position":685400},"target":"snapchain::connectors::fname"}
{"timestamp":"2024-11-02T01:22:35.752016Z","level":"DEBUG","fields":{"message":"fetching transfers","url":"https://fnames.farcaster.xyz/transfers?from_id=685500"},"target":"snapchain::connectors::fname"}
{"timestamp":"2024-11-02T01:22:36.281028Z","level":"INFO","fields":{"message":"found new transfers","count":100,"position":685500},"target":"snapchain::connectors::fname"}
{"timestamp":"2024-11-02T01:22:36.281093Z","level":"DEBUG","fields":{"message":"fetching transfers","url":"https://fnames.farcaster.xyz/transfers?from_id=685600"},"target":"snapchain::connectors::fname"}
{"timestamp":"2024-11-02T01:22:36.789052Z","level":"INFO","fields":{"message":"found new transfers","count":79,"position":685600},"target":"snapchain::connectors::fname"}
{"timestamp":"2024-11-02T01:22:36.789202Z","level":"DEBUG","fields":{"message":"fetching transfers","url":"https://fnames.farcaster.xyz/transfers?from_id=685679"},"target":"snapchain::connectors::fname"}
```

Currently I've hardcoded the start point at ID 685400 so that we exercise the code but don't flood the server.

The logic is: fetch without delay as long as non-error, non-empty responses come in. In the case of errors or empty responses we'll sleep for 5s.

The fetcher may exit with an error if it detects a non-incrementing sequence number.

In a follow on version I'll:
- add an enable/disable toggle
- add a config value to set the start ID.

### DIFF
--- a/src/connectors/fname/mod.rs
+++ b/src/connectors/fname/mod.rs
@@ -1,9 +1,7 @@
 use tokio::time::{sleep, Duration};
 use serde::Deserialize;
 use thiserror::Error;
-
-
-// TODO: use actual logger
+use tracing::{debug, info, warn, span, Level, Span};
 
 
 #[derive(Deserialize, Debug)]
@@ -30,6 +28,7 @@ pub enum FnamesError {
     NonSequentialIds,
 }
 
+
 #[derive(Error, Debug)]
 enum FetchError {
     #[error("non-sequential IDs found")]
@@ -44,43 +43,42 @@ enum FetchError {
 
 
 pub struct Fetcher {
-    count: u64,
+    position: u64,
     transfers: Vec<Transfer>,
 }
 
+
 impl Fetcher {
-    pub fn new(initial_count: u64) -> Self {
+    pub fn new(start_from: u64) -> Self {
         Fetcher {
-            count: initial_count,
+            position: start_from,
             transfers: vec![],
         }
     }
 
     async fn fetch(&mut self) -> Result<u64, FetchError> {
-        let mut count = self.count;
-
         loop {
-            let url = format!("https://fnames.farcaster.xyz/transfers?from_id={}", count);
-            println!("fname: fetching url: {}", url);
+            let url = format!("https://fnames.farcaster.xyz/transfers?from_id={}", self.position);
+            debug!(%url, "fetching transfers");
 
             let response = reqwest::get(&url)
                 .await?
                 .json::<TransfersData>()
                 .await?;
 
-            let transfers_count = response.transfers.len();
+            let count = response.transfers.len();
 
-            if transfers_count == 0 {
-                return Ok(count);
+            if count == 0 {
+                return Ok(self.position);
             }
 
-            println!("fname: found {} new transfers", transfers_count);
+            info!(count, position=self.position, "found new transfers");
 
             for t in response.transfers {
-                if t.id <= count {
+                if t.id <= self.position {
                     return Err(FetchError::NonSequentialIds);
                 }
-                count = t.id;
+                self.position = t.id;
                 self.transfers.push(t); // Just store these for now, we'll use them later
             }
         }
@@ -89,13 +87,13 @@ impl Fetcher {
     pub async fn run(&mut self) -> FnamesError {
         loop {
             match self.fetch().await {
-                Ok(new_count) => self.count = new_count,
+                Ok(new_count) => self.position = new_count,
                 Err(e) => match e {
                     FetchError::NonSequentialIds => {
                         return FnamesError::NonSequentialIds;
                     }
                     FetchError::Reqwest(request_error) => {
-                        println!("fname: reqwest error fetching: {}", request_error);
+                        warn!(error = %request_error, "reqwest error fetching transfers");
                     }
                     FetchError::NoNewIDs => {} // just sleep and retry
                 },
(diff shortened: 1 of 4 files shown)


src/main.rs
```diff
@@ -46,9 +49,12 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
 
     println!("SnapchainService (ID: {}) listening on {}", args.id, addr);
 
-    let _ = tracing_subscriber::fmt()
-        .with_env_filter(EnvFilter::from_default_env())
-        .try_init();
+    let env_filter = EnvFilter::try_from_default_env().unwrap_or_else(|_| EnvFilter::new("info"));
+
+    tracing_subscriber::fmt()
+        .json()
```
sanjayprabhu, line 55: JSON is important for prod, but we might want to use simple text for the dev env or until we're ready for testnet. We're going to be staring at a lot of log lines until then, and it's worth it to make it easy to read.

src/main.rs
```diff
@@ -69,6 +75,15 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
         println!("Gossip Stopped");
     });
 
+    // TODO: config to enable/disable running this fetcher
+    // TODO: add start_from as a parameter
+    let mut fetcher = Fetcher::new(685400u64);
```
sanjayprabhu, line 80: Another option is to just fetch the first 100 or so transfers or only do it for our fids. Main benefit is the data is going to be bounded and consistent across runs.

src/connectors/fname/mod.rs
```diff
@@ -0,0 +1,105 @@
...
+        }
+    }
+
+    pub async fn run(&mut self) -> FnamesError {
+        loop {
+            match self.fetch().await {
+                Ok(new_count) => self.position = new_count,
+                Err(e) => match e {
+                    FetchError::NonSequentialIds => {
+                        return FnamesError::NonSequentialIds;
```
sanjayprabhu, line 93: Can we use the same type everywhere?

Also, even in this scenario, we don't want the fnames fetcher to stop. Because that would mean the chain cannot ingest new dependent messages or it might require a restart of all the nodes. As a general principle, the nodes should prioritize uptime/liveness. 

This would also affect consensus (and cause chain halts) because if only certain validators see this error and stop fetching fnames, then they cannot validate fnames included by other validators that did not encounter this.

In practice, this error is unlikely to ever happen so it's not a big deal, but it's worth thinking through these things because these "consensus critical" issues are going to be the biggest source of pain.

src/connectors/fname/mod.rs
```diff
@@ -0,0 +1,105 @@
...
+                self.position = t.id;
+                self.transfers.push(t); // Just store these for now, we'll use them later
+            }
+        }
+    }
+
+    pub async fn run(&mut self) -> FnamesError {
+        loop {
+            match self.fetch().await {
+                Ok(new_count) => self.position = new_count,
```
sanjayprabhu, line 90: We're updating the position in both fetch and here.  Should we have fetch be a pure function and update the state here?

sanjayprabhu (commented):
Nice

src/main.rs
suurkivi: Interesting, perhaps both `from` and `to` parameters? I guess these will be temporary.

If we place an upper bound though, we won't be exercising the codepath that relaxes into the 5s polling delay that discovers new transfers. Although perhaps that's better handled by test cases (which we also don't have).

To get consistency between runs we might need a persistence layer as well, as a slow network may not fetch the same set of IDs before we start asking for them.

I'm tempted to defer the decision until we have a config system and things are further along.

src/connectors/fname/mod.rs
suurkivi: Yes, definitely worth thinking through!

With above context, agree that retrying is probably better. If we got non-sequential IDs, it probably means that something went really off the rails, and if it persists we'll probably break consensus and eventually halt production. But perhaps something resolves on the fname server in a retry and we're back in business.

> Can we use the same type everywhere?

Did you mean the fact that there was both NonSequentialIds in both FetchError and FnamesError? I'm not sure, this is my lack of rust experience shining through. I'd like to know what the best solution here is. This was the least ugly configuration I could come up with. But also since we'll be retrying instead of returning FnamesError, this issue will no longer exist.

src/connectors/fname/mod.rs
suurkivi: Yes, this was leftover from a refactor.

src/main.rs
suurkivi: Sounds good, let's add an config switch for logging format as soon as we have a config system.

sanjayprabhu (approved):

PR #6, opened 2024-11-04
config system

suurkivi:
Configuration system.

This allows configurations in TOML files:

```
log_format = "json"
id = 1

[fnames]
disable = false
stop_at = 9999999999
start_from = 0
url = "https://fnames.farcaster.xyz/transfers"
```

You specify the file with e.g. `cargo run -- --config-path node1.toml`

Currently possible to run without specifying any config file, but you'd have to pass the node id, e.g. `cargo run -- --id 1`

Environment variables take precedence over config files.

```
SNAPCHAIN_ID=1  cargo run --
````

or

```
SNAPCHAIN_FNAMES__URL=http://localhost:8000/transfers cargo run -- --id 1
```

Note the use of both single and double underscores to reference nested fields.

I don't love how much manual effort is required to add each command line flag (see the lengthy comment in code), but I couldn't see an obvious way to improve this using either `config` or `figment`. Open to ideas here but will likely defer to a future PR in any case.

I think there's an interesting discussion about where to store defaults for components. Do components define their own default config inside the module itself, or is there a centralized default config for the whole app?

This also updates the fnames fetcher to use the new config, and also add a few config options.  I'm happy to split into a separate PR if desired.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,7 +32,12 @@ blake3 = "1.4.1"
 tracing = "0.1.40"
 thiserror = "1.0.66"
 reqwest = { version = "0.12.9", features = ["json"] }
+figment = { version = "0.10.19", features = ["env", "toml"] }
 
 
 [build-dependencies]
 tonic-build = "0.9.2"
+
+[dev-dependencies]
+serial_test = "3.1.1"
+tempfile = "3.13.0"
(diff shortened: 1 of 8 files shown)


sanjayprabhu (commented):
Nice. We should also set up CI on github now that we have our first test.

sanjayprabhu (approved):
Worth checking in a sample config file

PR #8, opened 2024-11-05
grpc server scaffolding

suurkivi:
grpc server scaffolding

For now, run on port 50060 + <id>

```
grpcurl -plaintext -import-path src/proto -proto src/proto/rpc.proto -d '{}' 127.0.0.1:50061 rpc.SnapchainService/SubmitMessage
ERROR:
  Code: Unimplemented
  Message: not implemented
```


src/main.rs
```diff
@@ -85,6 +103,25 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+            .serve(grpc_socket_addr)
+            .await;
+
+        let msg = "grpc server stopped";
+        match resp {
+            Ok(()) => error!(msg),
+            Err(e) => error!(error = ?e, "{}", msg),
+        }
+
+        shutdown_tx.send(()).await.ok();
```
sanjayprabhu, line 122: When will this be hit?

src/main.rs
```diff
@@ -138,7 +180,7 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
                         }),
                         nonce: tick_count as u64,   // Need the nonce to avoid the gossip duplicate message check
                     };
-                    println!("Registering validator with nonce: {}", register_validator.nonce);
+                    info!("Registering validator with nonce: {}", register_validator.nonce);
```
sanjayprabhu, line 183: Thanks, I'll swap the remaining printlns with tracing in my next pr

src/server.rs
```diff
@@ -0,0 +1,34 @@
...
+}
+
+use rpc::snapchain_service_server::{SnapchainService, SnapchainServiceServer};
+use message::{Message};
+
+#[derive(Default)]
+pub struct MySnapchainService;
+
+#[tonic::async_trait]
+impl SnapchainService for MySnapchainService {
```
sanjayprabhu, line 28: We should put this into `network/`

build.rs
```diff
@@ -1,10 +1,15 @@
...
     // Custom type attributes required for malachite
     builder = builder.type_attribute("snapchain.ShardHash", "#[derive(Eq, PartialOrd, Ord)]");
 
-    builder.compile(&["src/proto/blocks.proto"], &["src/proto"])?;
+    builder.compile(&[
+        "src/proto/blocks.proto",
+        "src/proto/rpc.proto",
+        "src/proto/message.proto",
+        "src/proto/username_proof.proto",
+    ], &["src/proto"])?;
```
sanjayprabhu, line 12: Would be nice to autodiscover all the files, but not a priority

sanjayprabhu (approved):

src/main.rs
suurkivi: I guess for any reason the grpc server stops running. I'm not sure the full set of cases, but one common one is when the grpc server address/port is already in use.

build.rs
suurkivi: Added a TODO

Rufusfavour: Good to see this implemented as proposed in issue 5 some days ago
https://github.com/farcasterxyz/snapchain-v0/issues/5#

PR #10, opened 2024-11-06
start consuming on chain events

aditiharini:
Read onchain events and store them in memory. 

Tested with: 
`SNAPCHAIN_ONCHAIN_EVENTS__RPC_URL=<alchemy rpc url> cargo run -- --id 1`
The on chain events consumer logs every time we receive a new onchain event. I tested further by dumping the state of the module whenever there was a new on chain event. 

Example log: 
```
2024-11-06T16:26:25.880328Z  INFO snapchain::connectors::onchain_events: Processed onchain event Event {
    chain_id: 10,
    block_number: 127655803,
    block_hash: 0x72b93f463736aaaf79ae3ddb0583677201150fd312138620a50502fe72b69c16,
    block_timestamp: 1730910383,
    log_index: 86,
    fid: 874469,
    tx_index: 40,
    version: 0,
    event_type: IdRegister(
        Register {
            to: 0x9a6954bf025a570f987f7315a46121a7200f9770,
            recovery_address: 0x00000000fcb080a4d6c39a9354da9eb9bc104cd7,
        },
    ),
}
2024-11-06T16:26:25.968555Z  INFO snapchain::connectors::onchain_events: Processed onchain event Event {
    chain_id: 10,
    block_number: 127655803,
    block_hash: 0x72b93f463736aaaf79ae3ddb0583677201150fd312138620a50502fe72b69c16,
    block_timestamp: 1730910383,
    log_index: 87,
    fid: 874469,
    tx_index: 40,
    version: 0,
    event_type: StorageRent(
        StorageRentEvent {
            payer: 0x00000000fc25870c6ed6b6c7e41fb078b7656f69,
            units: 1,
            expiry: 1762446383,
        },
    ),
}
2024-11-06T16:26:26.048178Z  INFO snapchain::connectors::onchain_events: Processed onchain event Event {
    chain_id: 10,
    block_number: 127655803,
    block_hash: 0x72b93f463736aaaf79ae3ddb0583677201150fd312138620a50502fe72b69c16,
    block_timestamp: 1730910383,
    log_index: 88,
    fid: 874469,
    tx_index: 40,
    version: 0,
    event_type: Signer(
        Add {
            key: 0x4e0370582dc260309de8ffc341db721e143b03c6619e2b11f789922f793cf00b,
            key_type: 1,
            metadata: 0x000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000023c000000000000000000000000002ef790dd7993a35fd847c053eddae940d055596000000000000000000000000000000000000000000000000000000000000008000000000000000000000000000000000000000000000000000000000672b9b030000000000000000000000000000000000000000000000000000000000000041e835777acf64d852ef104f19491c5e9b5be6177ff629713200ec122019e726502fcb690737217078d2e3b1f800359c8793275fb06f74038d083670c5a0e4bcc71b00000000000000000000000000000000000000000000000000000000000000,
            metadata_type: 1,
        },
    ),
}
```

### DIFF
--- a/src/connectors/onchain_events/mod.rs
+++ b/src/connectors/onchain_events/mod.rs
@@ -346,12 +346,12 @@ impl Subscriber {
                 let result = self.process_log(&event).await;
                 if result.is_err() {
                     error!(
-                        "Error processing on chain event. Error: {:#?}. Event: {:#?}",
+                        "Error processing onchain event. Error: {:#?}. Event: {:#?}",
                         result.err(),
                         event,
                     )
                 } else {
-                    debug!("Processed event {:#?}", event);
+                    debug!("Processed onchain event {:#?}", event);
                 }
             }
         }
--- a/src/connectors/onchain_events/mod.rs
+++ b/src/connectors/onchain_events/mod.rs
@@ -40,7 +40,7 @@ static KEY_REGISTRY: Address = address!("00000000Fc1237824fb747aBDE0FF18990E59b7
 
 static ID_REGISTRY: Address = address!("00000000Fc6c5F01Fc30151999387Bb99A9f489b");
 
-static CHAIN_ID: u64 = 10;
+static CHAIN_ID: u64 = 10; // OP mainnet
 const RENT_EXPIRY_IN_SECONDS: u64 = 365 * 24 * 60 * 60; // One year
 
 #[derive(Debug, Serialize, Deserialize)]
--- a/src/connectors/onchain_events/mod.rs
+++ b/src/connectors/onchain_events/mod.rs
@@ -46,14 +46,12 @@ const RENT_EXPIRY_IN_SECONDS: u64 = 365 * 24 * 60 * 60; // One year
 #[derive(Debug, Serialize, Deserialize)]
 pub struct Config {
     pub rpc_url: String,
-    pub disable: bool,
 }
 
 impl Default for Config {
     fn default() -> Config {
         return Config {
-            rpc_url: "must specify rpc url in config".to_string(),
-            disable: false,
+            rpc_url: String::new(),
         };
     }
 }
@@ -68,6 +66,9 @@ pub enum SubscribeError {
 
     #[error(transparent)]
     UnableToParseLog(#[from] alloy_sol_types::Error),
+
+    #[error("Empty rpc url")]
+    EmptyRpcUrl,
 
     #[error("Log missing block hash")]
     LogMissingBlockHash,
@@ -157,15 +158,18 @@ pub struct Subscriber {
 // TODO(aditi): Wait for 1 confirmation before "committing" an onchain event.
 impl Subscriber {
     pub fn new(config: Config) -> Result<Subscriber, SubscribeError> {
+        if config.rpc_url.is_empty() {
+            return Err(SubscribeError::EmptyRpcUrl);
+        }
         let url = config.rpc_url.parse()?;
         let provider = ProviderBuilder::new().on_http(url);
-        return Ok(Subscriber {
+        Ok(Subscriber {
             provider,
             onchain_events_by_block: HashMap::new(),
-        });
+        })
     }
 
-    pub fn add_onchain_event(
+    fn add_onchain_event(
         &mut self,
         fid: u64,
         block_number: u64,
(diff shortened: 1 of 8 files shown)


src/connectors/onchain_events/mod.rs
```diff
@@ -0,0 +1,361 @@
...
+pub struct Config {
+    pub rpc_url: String,
+    pub disable: bool,
+}
+
+impl Default for Config {
+    fn default() -> Config {
+        return Config {
+            rpc_url: "must specify rpc url in config".to_string(),
+            disable: false,
```
suurkivi, line 56: I wonder if we should disable by default (at least for now) so that snapchain can run without needing to set the rpc_url?

In the future, is it possible to point to a testnet (or some other thing) that doesn't need an API key?

What do we do for hubs? Maybe it's totally normal to just need your own API key.

src/connectors/onchain_events/mod.rs
```diff
@@ -0,0 +1,361 @@
...
+#[derive(Debug, Serialize, Deserialize)]
+pub struct Config {
+    pub rpc_url: String,
+    pub disable: bool,
+}
+
+impl Default for Config {
+    fn default() -> Config {
+        return Config {
+            rpc_url: "must specify rpc url in config".to_string(),
```
suurkivi, line 55: I noticed when this is set to a invalid url, the program exits with:

Error: UnableToParseUrl(RelativeUrlWithoutBase)

and a little hard to debug.

As a separate idea, maybe the default value can be empty (or make it optional) and then do a check at startup (but only when the component is enabled)

src/connectors/onchain_events/mod.rs
```diff
@@ -0,0 +1,361 @@
...
+    pub fn new(config: Config) -> Result<Subscriber, SubscribeError> {
+        let url = config.rpc_url.parse()?;
+        let provider = ProviderBuilder::new().on_http(url);
+        return Ok(Subscriber {
+            provider,
+            onchain_events_by_block: HashMap::new(),
+        });
+    }
+
+    pub fn add_onchain_event(
```
suurkivi, line 168: nit: a few of these can be private

src/connectors/onchain_events/mod.rs
```diff
@@ -0,0 +1,361 @@
...
+
+    pub async fn run(&mut self) -> Result<(), SubscribeError> {
+        // Subscribe to new events starting from now.
+        let filter = Filter::new().address(vec![STORAGE_REGISTRY, KEY_REGISTRY, ID_REGISTRY]);
+        let subscription = self.provider.watch_logs(&filter).await?;
+        let mut stream = subscription.into_stream();
+        while let Some(events) = stream.next().await {
+            for event in events {
+                let result = self.process_log(&event).await;
+                if result.is_err() {
```
suurkivi, line 348: nit: I (guess!) the idiomatic way would be to use `match` here. But I've be writing rust for only a few days! And I don't know if we have any established code conventions.

src/connectors/onchain_events/mod.rs
```diff
@@ -0,0 +1,361 @@
...
+            tx_index,
+            event_type,
+            chain_id: CHAIN_ID,
+            version: 0,
+        };
+        let events = self.onchain_events_by_block.get_mut(&block_number);
+        match events {
+            None => {
+                self.onchain_events_by_block
+                    .insert(block_number, vec![event]);
```
suurkivi, line 193: Is this setting the key to the empty vec if it doesn't exist? Do we also have to push the event in this case? If not, perhaps a comment to explain?

src/connectors/onchain_events/mod.rs
```diff
@@ -0,0 +1,361 @@
...
+                );
+                return Ok(());
+            }
+            Some(&KeyRegistryAbi::Migrated::SIGNATURE_HASH) => {
+                let KeyRegistryAbi::Migrated { keysMigratedAt } = event.log_decode()?.inner.data;
+                let migrated_at = Uint::to::<u64>(&keysMigratedAt);
+                add_event(0, EventType::SignerMigrated { migrated_at });
+                return Ok(());
+            }
+            _ => return Ok(()),
```
suurkivi, line 336: What events might captured by this path? Is it possible to cover all cases? (my guess is no, but not sure).

Probably worth logging/warning something here? Or would it be too noisy?

src/connectors/onchain_events/mod.rs
aditiharini: In hubs we currently require everyone to provide an rpc url in the startup command. 

Happy to disable by default for a while.

src/connectors/onchain_events/mod.rs
aditiharini: I might be misunderstanding this comment, but we do insert the event. The vec is initialized with the event inside of it.

src/connectors/onchain_events/mod.rs
aditiharini: There are a bunch of events we don't process. I'm not sure how often they actually show up, though. IMO it seems ok to just leave this codepath without logging--- we don't actually need to know if there are other events being emitted.

src/connectors/onchain_events/mod.rs
suurkivi: Yes, missed that thanks. Still getting used to reading rust.

suurkivi (approved):
nice

PR #12, opened 2024-11-06
WIP: FC message channels

suurkivi:


src/consensus/consensus.rs
```diff
@@ -176,7 +179,17 @@ impl ShardValidator {
...
-            shard_chunks: vec![],
+            shard_chunks: vec![ShardChunk {
+                hash: vec![0, 1, 2, 3],
+                header: None,
+                votes: None,
+                transactions: vec![Transaction {
+                    fid: 1234,
+                    account_root: vec![5, 5, 6, 6],
+                    system_messages: vec![],
+                    user_messages: messages.into_iter().map(|_| UserMessage {}).collect(), // TODO
```
sanjayprabhu, line 190: `UserMessage` can just be `FCMessage`, so just need to provide it `messages`

PR #14, opened 2024-11-07
use lib

suurkivi:
This is the first step in being able to import protobufs from a centralized place.

[lib]
name = "snapchain"
path = "src/lib.rs"

Pretty sure this is the convention and having src/main.rs is causing all kinds of mischief.


sanjayprabhu (approved):
Thanks, sorry about the inconvenience. I had a mental note to update this to lib.rs after your PR but I didn't realize it would cause these issues.

PR #16, opened 2024-11-08
add support for persisting blocks

aditiharini:
Introduce rocksdb to store blocks. 

Each node should get its own rocksdb directory on startup: `SNAPCHAIN_ROCKSDB_DIR=".rocks5" cargo run -- --id 5`

The rocksdb interfacing code is mostly copied over from hubble. The main changes are:
- Indexing: We were indexing messages by fid, timestamp. I changed to index by shard_id, block number. When I say “index”, I mean how we construct the prefixes for the strings we insert into rocksdb.
- Start/stop iterators: Our code for wrapper code iterating over elements in the db didn’t take a start and stop point for the iteration. It took just a single prefix and would iterate all the elements with that prefix. I modified the wrapper code to support this.

Running 
`cargo run --bin follow-blocks` and submitting a message works: 
```
...
Block 15
Block 16
Block 17
Block 18
Block 19
Block 20
Block 21
 - adc13047f02b979415c140ee08ac9fb644d72882
Block 22
Block 23
Block 24
Block 25
Block 26
...
```


src/cfg.rs
```diff
@@ -27,6 +28,7 @@ impl Default for Config {
             onchain_events: connectors::onchain_events::Config::default(),
             consensus: consensus::consensus::Config::default(),
             rpc_address: "0.0.0.0:3383".to_string(),
+            rocksdb_dir: ".rocks".to_string(),
```
suurkivi, line 31: should we modify setup.rs to create a different directory for each node?

src/consensus/consensus.rs
```diff
@@ -314,56 +321,61 @@ pub enum BlockProposerError {
 
     #[error(transparent)]
     RpcResponseError(#[from] tonic::Status),
+
+    #[error(transparent)]
+    BlockStorageError(#[from] BlockStorageError),
 }
-#[derive(Default)]
+#[derive(Default, Clone)]
```
suurkivi, line 328: Perhaps in a follow on PR we could move this to a separate file (and maybe some other big structs in here?)

src/bin/setup.rs
```diff
@@ -23,6 +23,7 @@ async fn main() {
             r#"
 id = {id}
 rpc_address=0.0.0.0:{rpc_port}
+rocksdb_dir=.rocks{i}
```
sanjayprabhu, line 26: I think this should be `nodes/id/1/.rocks`. All data for a particular node should go into the node directory.

sanjayprabhu: Let us also delete the existing .rocks directory each time we run `setup` (can customize this later, but for now, I think we don't want to preserve any state across runs)

src/cfg.rs
sanjayprabhu, line 31: Add this to .gitignore?

src/storage/db/rocksdb.rs
```diff
@@ -0,0 +1,541 @@
...
+    pub reverse: bool,
+}
+
+impl std::fmt::Debug for RocksDB {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("RocksDB").field("path", &self.path).finish()
+    }
+}
+
+impl RocksDB {
```
sanjayprabhu, line 79: Is this copied as is from the hub? Any notable changes?

src/storage/store/block.rs
```diff
@@ -0,0 +1,133 @@
...
+}
+
+pub const PAGE_SIZE_MAX: usize = 10_000;
+/** A page of messages returned from various APIs */
+pub struct BlockPage {
+    pub blocks: Vec<Block>,
+    pub next_page_token: Option<Vec<u8>>,
+}
+
+fn make_primary_key(shard_index: u32, block_number: u64) -> Vec<u8> {
```
sanjayprabhu, line 30: `make_shard_key` or `make_block_key`?

src/storage/store/block.rs
```diff
@@ -0,0 +1,133 @@
...
+) -> Result<BlockPage, BlockStorageError> {
+    let start_primary_key = make_primary_key(shard_index, start_block_number);
+    let stop_prefix =
+        stop_block_number.map(|block_number| make_primary_key(shard_index, block_number));
+
+    get_block_page_by_prefix(db, page_options, Some(start_primary_key), stop_prefix)
+}
+
+pub fn put_block(db: &RocksDB, block: Block) -> Result<(), BlockStorageError> {
+    let mut txn = db.txn();
```
sanjayprabhu, line 120: This is fine for now, but this is the what will change when we implement the transaction model. We should only commit to the db when a block if confirmed. And all db changes should be atomic

tests/consensus_test.rs
```diff
@@ -239,6 +268,7 @@ impl TestNetwork {
     pub fn stop(&self) {
         for node in self.nodes.iter() {
             node.stop();
+            node.db.destroy().unwrap();
```
sanjayprabhu, line 271: Does this get called correctly if tests fail/panic?

Cargo.toml
```diff
@@ -40,9 +39,9 @@ url = "2.5.3"
 alloy-transport = "0.5.4"
 alloy-sol-types = "0.8.11"
 ed25519-dalek = "2.1.1"
-
 pre-commit = "0.5.2"
-
+rocksdb = {git = "https://github.com/rust-rocksdb/rust-rocksdb.git", features=["multi-threaded-cf"]}
```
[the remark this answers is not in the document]
aditiharini: There was an issue with rocksdb startup in the pinned version. This is the fix but it hasn't been captured in a new version: https://github.com/rust-rocksdb/rust-rocksdb/commit/1cf906dc4087f06631820f13855e6b27bd21b972. 

It's possible that playing around with rust toolchains might solve this but I felt like doing this for now is fine.

src/storage/db/rocksdb.rs
aditiharini: Added a note on material changes in the description.

src/storage/store/block.rs
aditiharini: Sounds good, added a comment.

Cargo.toml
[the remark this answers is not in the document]
aditiharini: Pinned the rocksdb repo to the rev with the fix we need.

tests/consensus_test.rs
aditiharini: Ah no it doesn't. I implemented the `Drop` trait on `NodeForTest` to handle cleanup. Might be better to impl this closer to the source (e.g. on the rocksdb type) but this seems like it does the job for now.

src/storage/store/block.rs
```diff
@@ -0,0 +1,134 @@
...
+}
+
+fn make_block_key(shard_index: u32, block_number: u64) -> Vec<u8> {
+    let mut key = [0u8; 12];
+    // Store the shard index as big-endian in the first 4 bytes
+    key[0..4].copy_from_slice(&shard_index.to_be_bytes());
+    // Store the block number in the remaining 8 bytes
+    key[4..12].copy_from_slice(&block_number.to_be_bytes());
+
+    key.to_vec()
```
sanjayprabhu, line 37: We should have a global prefix for the type of data we're storing, similar to the Hub's RootPrefix https://github.com/farcasterxyz/hub-monorepo/blob/7b3748906efa1a5b8f9b2755c2289d0fc8904ff8/apps/hubble/src/storage/db/types.ts#L23

Otherwise we're going to have prefix collisions when we store account state.

sanjayprabhu (approved):

src/storage/store/block.rs
aditiharini: Ah good point, added.

PR #18, opened 2024-11-08
setup block sync api

aditiharini:
Set up RPC and wiring for block sync.

### DIFF
--- a/src/consensus/consensus.rs
+++ b/src/consensus/consensus.rs
@@ -329,9 +329,8 @@ pub struct BlockProposer {
     proposed_blocks: BTreeMap<ShardHash, FullProposal>,
     shard_decision_rx: RxDecision,
     num_shards: u32,
-    tx_decision: Option<TxDecision>,
+    block_tx: Option<mpsc::Sender<Block>>,
     block_store: BlockStore,
-    new_block_tx: mpsc::Sender<Block>,
 }
 
 impl BlockProposer {
@@ -340,8 +339,7 @@ impl BlockProposer {
         shard_id: SnapchainShard,
         shard_decision_rx: RxDecision,
         num_shards: u32,
-        tx_decision: Option<TxDecision>,
-        new_block_tx: mpsc::Sender<Block>,
+        block_tx: Option<mpsc::Sender<Block>>,
     ) -> BlockProposer {
         BlockProposer {
             shard_id,
@@ -350,8 +348,7 @@ impl BlockProposer {
             proposed_blocks: BTreeMap::new(),
             shard_decision_rx,
             num_shards,
-            tx_decision,
-            new_block_tx,
+            block_tx,
             block_store: BlockStore::new(),
         }
     }
@@ -405,12 +402,9 @@ impl BlockProposer {
     }
 
     async fn publish_new_block(&self, block: Block) {
-        match self.new_block_tx.send(block.clone()).await {
-            Err(err) => {
-                error!("Unable to send block {:#?}", block);
-            }
-            Ok(()) => {}
-        };
+        if let Some(block_tx) = &self.block_tx {
+            let _ = block_tx.send(block.clone()).await;
+        }
     }
 
     pub async fn register_validator(
@@ -520,9 +514,6 @@ impl Proposer for BlockProposer {
 
     async fn decide(&mut self, _height: Height, _round: Round, value: ShardHash) {
         if let Some(proposal) = self.proposed_blocks.get(&value) {
-            if let Some(tx_decision) = &self.tx_decision {
-                let _ = tx_decision.send(proposal.clone()).await;
-            }
             self.publish_new_block(proposal.block().unwrap());
             self.blocks.push(proposal.block().unwrap());
             self.proposed_blocks.remove(&value);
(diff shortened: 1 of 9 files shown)


src/main.rs
```diff
@@ -150,9 +136,15 @@ async fn main() -> Result<(), Box<dyn Error>> {
     };
 
     let ctx = SnapchainValidatorContext::new(keypair.clone());
-    let block_proposer = BlockProposer::new(validator_address.clone(), shard.clone());
-    let shard_validator =
-        ShardValidator::new(validator_address.clone(), Some(block_proposer), None);
+    let block_proposer = Arc::new(Mutex::new(BlockProposer::new(
+        validator_address.clone(),
+        shard.clone(),
```
sanjayprabhu, line 141: We shouldn't share the proposer across threads. Whenever you end up requiring Arc/Mutex, we should think rethink how the dataflow works.


We should be using Message passing for concurrency instead of sharing memory: https://doc.rust-lang.org/book/ch16-02-message-passing.html

Take a look at how the block proposer is listening for confirmed shard chunks https://github.com/farcasterxyz/snapchain-v0/blob/2aa2393bc3b073fed82c2d5e2f73693ab0be59c2/src/consensus/consensus.rs#L292.


When @suurkivi is back we should talk about a higher level framework that will make this easier. I'm not a huge fan of ractor/actor model (it's very hard to follow the control flow), but it feels like we might end up rebuilding a lot of it if we try to do it from scratch.


In this case, I might do something like:
1. create a channel for the block proposer to communicate to the grpc server.
2. In register validator, send the highest height the validator has. And in consensus actor, when it receives the message, call proposer.is_ready_for_height (and start height should only be called when the proposer is ready)
3. If the proposer is missing blocks, send a message on the grpc server tx with a oneshot rx to receive the reply. For now you can just await on the rx, but we should push this to the background
4. When the gap is filled, it can return true and StartHeight

src/proto/blocks.proto
```diff
@@ -8,6 +8,7 @@ message Validator {
   bytes signer = 2;
   string rpc_address = 3;
   uint32 shard_index = 4;
+  uint64 max_known_block_number = 5;
```
sanjayprabhu, line 11: Something like `current_height` is a better name

src/main.rs
```diff
@@ -103,8 +106,35 @@ async fn main() -> Result<(), Box<dyn Error>> {
...
+    // Use the new non-global metrics registry when we upgrade to newer version of malachite
+    let _ = Metrics::register(registry);
+
+    let (rpc_block_tx, mut rpc_block_rx) = mpsc::channel::<Block>(100);
+    let (main_block_tx, mut main_block_rx) = mpsc::channel::<Block>(100);
+
+    let mut block_store = BlockStore::new();
+    while let Some(block) = main_block_rx.recv().await {
+        block_store.put_block(block)
+    }
```
sanjayprabhu, line 119: Why do we need this? Who's publishing at this point?

src/consensus/consensus.rs
```diff
@@ -336,6 +413,55 @@ impl BlockProposer {
         }
         confirmed_shard_chunks
     }
+
+    async fn publish_new_block(&self, block: Block) {
+        for tx in &self.block_tx {
+            let _ = tx.send(block.clone()).await;
+        }
```
sanjayprabhu, line 420: A better option might be to use a broadcast channel for this https://docs.rs/tokio/latest/tokio/sync/broadcast/index.html

src/consensus/consensus.rs
```diff
@@ -680,6 +805,15 @@ impl Consensus {
                 // let total_peers = state.consensus.driver.validator_set().count() - 1;
 
                 // println!("Connected to {connected_peers}/{total_peers} peers");
+                match &mut state.shard_validator.block_proposer {
+                    None => {}
+                    Some(block_proposer) => {
+                        match block_proposer.register_validator(&validator).await {
```
sanjayprabhu, line 811: Let's have the shard validator forward the call to the block proposer. This is looking too deep into the validator.

src/consensus/consensus.rs
```diff
@@ -936,10 +1070,15 @@ impl Actor for Consensus {
     ) -> Result<(), ActorProcessingErr> {
         state.timers.cancel_all();
         // Add ourselves to the validator set
+        let max_known_block_number = match &state.shard_validator.block_proposer {
+            None => 0,
+            Some(block_proposer) => block_proposer.block_store.max_block_number(),
+        };
```
sanjayprabhu, line 1076: This should be something like `shard_validator.get_current_height(shard_id)`

src/core/types.rs
```diff
@@ -266,19 +266,22 @@ pub struct SnapchainValidator {
     pub address: Address,
     pub public_key: PublicKey,
     pub rpc_address: Option<String>,
+    pub max_known_block_number: u64,
```
sanjayprabhu, line 269: current_height

src/consensus/consensus.rs
aditiharini: I ended up using a lock as a temporary solution til we have the db to avoid the messiness around channels. Once we have the db in, we won't need a broadcast channel.

src/main.rs
aditiharini: Removed this. This code wouldn't work and not needed anymore because of the lock.

PR #20, opened 2024-11-09
add rpc address to RegisterValidator

aditiharini:
We need to be able to reach peers via rpc for block sync, so add the rpc address to the broadcasted info about each validator.

### DIFF
--- a/src/cfg.rs
+++ b/src/cfg.rs
@@ -26,7 +26,7 @@ impl Default for Config {
             fnames: connectors::fname::Config::default(),
             onchain_events: connectors::onchain_events::Config::default(),
             consensus: consensus::consensus::Config::default(),
-            rpc_address: "0.0.0.0:2283".to_string(),
+            rpc_address: "0.0.0.0:3383".to_string(),
         }
     }
 }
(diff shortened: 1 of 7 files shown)


src/cfg.rs
```diff
@@ -25,6 +26,7 @@ impl Default for Config {
             fnames: connectors::fname::Config::default(),
             onchain_events: connectors::onchain_events::Config::default(),
             consensus: consensus::consensus::Config::default(),
+            rpc_address: "0.0.0.0:2283".to_string(),
```
sanjayprabhu, line 29: We should pick something like 3383. Also, update setup.rs to generate ports that don't collide for the different nodes.

src/main.rs
```diff
@@ -135,7 +135,11 @@ async fn main() -> Result<(), Box<dyn Error>> {
 
     let shard = SnapchainShard::new(0); // Single shard for now
     let validator_address = Address(keypair.public().to_bytes());
-    let validator = SnapchainValidator::new(shard.clone(), keypair.public().clone());
+    let validator = SnapchainValidator::new(
+        shard.clone(),
+        keypair.public().clone(),
+        Some(app_config.rpc_address.clone()),
```
sanjayprabhu, line 141: Should this be an option?

sanjayprabhu (approved):

src/main.rs
aditiharini: Made it an option because I didn't want to store an rpc address on the validator running inside this node.

PR #22, opened 2024-11-12
message routing

suurkivi:


tests/consensus_test.rs
```diff
@@ -114,8 +125,9 @@ impl NodeForTest {
 
 #[tokio::test]
 async fn test_basic_consensus() {
+    let env_filter = EnvFilter::try_from_default_env().unwrap_or_else(|_| EnvFilter::new("info"));
```
sanjayprabhu, line 128: Let's keep this to `warn`. We could disable entirely for tests, but I've found this useful for debugging.

src/node/snapchain_node.rs
```diff
@@ -7,19 +7,21 @@ use crate::core::types::{
...
-use std::collections::BTreeMap;
+use std::collections::{BTreeMap, HashMap};
 use tokio::sync::mpsc;
 use tracing::warn;
 
 const MAX_SHARDS: u32 = 3;
 
 pub struct SnapchainNode {
     pub consensus_actors: BTreeMap<u32, ActorRef<ConsensusMsg<SnapchainValidatorContext>>>,
+    pub shard_messages: HashMap<u32, mpsc::Sender<message::Message>>,
```
sanjayprabhu, line 24: I like suffixing tx/rx to channels, so it's clearer when reading the code.

src/consensus/consensus.rs
```diff
@@ -222,10 +234,23 @@ impl Proposer for ShardProposer {
             .as_bytes()
             .to_vec();
 
+        let it = iter::from_fn(|| self.messages_rx.try_recv().ok());
+        let user_messages: Vec<message::Message> = it.collect();
```
sanjayprabhu, line 238: Does this collect everything until the channel is empty? I think we should impose a max limit and also a timeout (use what's being passed in above).

PR #23, opened 2024-11-12
Fix multiple shards and clean up tests

sanjayprabhu:
Creating a node with 2 shards wasn't working because the block proposer was using the length of the `shard_ids` string as the num_nodes instead of the length of the vector.

Reduced duplication in tests and made it easier to spin up a network of nodes.

Remove re-implementation of Height and used Height prost.

### DIFF
--- a/src/consensus/consensus.rs
+++ b/src/consensus/consensus.rs
@@ -19,7 +19,7 @@ use malachite_metrics::Metrics;
 
 use crate::consensus::timers::{TimeoutElapsed, TimerScheduler};
 use crate::core::types::proto::ShardHash;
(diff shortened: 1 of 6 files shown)


src/consensus/consensus.rs
```diff
@@ -73,7 +73,18 @@ impl Config {
     }
 
     pub fn num_shards(&self) -> u32 {
-        self.shard_ids.len() as u32
+        self.shard_ids().len() as u32
```
sanjayprabhu, line 76: This was why it wasn't working for multiple shards it was using "1,2".len() instead of vec![1,2].len(). So the Block proposer was awaiting for a 3rd chunk that would never arrive.

PR #27, opened 2024-11-14
setup stats collection script

aditiharini:
Connect to a node and dump some stats at an interval. 

`cargo run --bin collect_stats`

```
Blocks produced : 4, Messages submitted: 10, Messages processed: 12, Avg time to confirmation (secs): 0.6666666666666666
Blocks produced : 3, Messages submitted: 10, Messages processed: 11, Avg time to confirmation (secs): 0.5454545454545454
Blocks produced : 4, Messages submitted: 10, Messages processed: 12, Avg time to confirmation (secs): 0.4166666666666667
Blocks produced : 3, Messages submitted: 10, Messages processed: 11, Avg time to confirmation (secs): 0.2727272727272727
Blocks produced : 4, Messages submitted: 10, Messages processed: 11, Avg time to confirmation (secs): 1.1818181818181819
Blocks produced : 3, Messages submitted: 10, Messages processed: 12, Avg time to confirmation (secs): 1.0
```


src/bin/collect_stats.rs
```diff
@@ -0,0 +1,88 @@
...
+        // Create a timer for block creation
+        let mut submit_message_timer = time::interval(SUBMIT_MESSAGE_INTERVAL);
+
+        loop {
+            let _ = submit_message_timer.tick().await;
+            let message = compose_message(
+                private_key.clone(),
+                6833,
+                RPC_ADDR.to_string(),
+                "For benchmarking",
```
suurkivi, line 41: Maybe we can append an incrementing number each message? When we start actually hashing these messages we probably want to ensure they hash to different values (although hashing to the same value would be a different kind of test).

aditiharini: Sounds good, made the change.

PR #28, opened 2024-11-14
bring in trie/ from github.com/farcasterxyz/hub-monorepo.git @ 15d0aec9ba7eb9fb20fef5a5edc02d71bc02f7d3

suurkivi:


src/trie/merkle_trie.rs
```diff
@@ -0,0 +1,707 @@
+use super::trie_node::{TrieNode, TIMESTAMP_LENGTH};
```
sanjayprabhu, line 1: Let's move this into src/storage/trie

PR #33, opened 2024-11-15
trie commit and rollback v0

suurkivi:
This is a little messy but it does a few things:

- hashes of messages are added to the trie during propose, then rolled back
- hashes are again added during commit, and we check that the hashes match before committing or (for not at least) panicking

Still a lot to do but POC somewhat validated.


aditiharini:
Where in the code does the rollback currently happen?

PR #34, opened 2024-11-15
Storage abstraction

sanjayprabhu:
Creates an Snapchain Engine struct that's responsible for validating and committing state changes to the db.

At a high level, I think the engine should be the only one that holds a reference to the DB, and it manages creating and committing/rolling back transactions. All the other stores should accept a transaction as a parameter and perform their db operations using that.

### DIFF
--- a/src/consensus/validator.rs
+++ b/src/consensus/validator.rs
@@ -53,14 +53,13 @@ impl ShardValidator {
         self.validator_set.count()
     }
 
-    pub fn get_current_height(&self) -> Result<u64, BlockProposerError> {
-        if let Some(proposer) = &self.block_proposer {
-            Ok(proposer.get_confirmed_height().block_number)
-        } else if let Some(proposer) = &self.shard_proposer {
-            Ok(proposer.get_confirmed_height().block_number)
-        } else {
-            panic!("No proposer set");
+    pub fn get_current_height(&self) -> u64 {
+        if let Some(p) = &self.block_proposer {
+            return p.get_confirmed_height().block_number;
+        } else if let Some(p) = &self.shard_proposer {
+            return p.get_confirmed_height().block_number;
         }
+        panic!("No proposer set on validator");
     }
 
     pub fn add_validator(&mut self, validator: SnapchainValidator) -> bool {
--- a/src/consensus/validator.rs
+++ b/src/consensus/validator.rs
@@ -54,12 +54,12 @@ impl ShardValidator {
     }
 
     pub fn get_current_height(&self) -> Result<u64, BlockProposerError> {
-        match &self.block_proposer {
-            None => Ok(0),
-            Some(block_proposer) => block_proposer
-                .block_store
-                .max_block_number(self.shard_id.shard_id())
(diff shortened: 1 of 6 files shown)


src/consensus/validator.rs
```diff
@@ -54,12 +54,12 @@ impl ShardValidator {
     }
 
     pub fn get_current_height(&self) -> Result<u64, BlockProposerError> {
-        match &self.block_proposer {
-            None => Ok(0),
-            Some(block_proposer) => block_proposer
-                .block_store
-                .max_block_number(self.shard_id.shard_id())
-                .map_err(|err| BlockProposerError::BlockStorageError(err)),
+        if let Some(proposer) = &self.block_proposer {
```
suurkivi, line 57: nit: this took me a lot longer to read than I might have expected (maybe I'm still too new to rust)

```rust
pub fn get_current_height(&self) -> Result<u64, BlockProposerError> {
    if let Some(p) = &self.block_proposer {
        return Ok(p.get_confirmed_height().block_number);
    }
    if let Some(p) = &self.shard_proposer {
        return Ok(p.get_confirmed_height().block_number);
    }
    Err(BlockProposerError::NoProposerSet)
}
```

I like return above else if / else.

I like `p` here, proposer is just too close to both block_proposer and shard_proposer, with so many `proposer`s it took a long time to realize the two cases were actually different. Instead of `p` I think `block_proposer` and `shard_proposer` could work too.

Feel free to ignore any of above.

src/storage/store/engine.rs
```diff
@@ -0,0 +1,121 @@
...
+        // Replay the state change
+        // If the state root does not match or any of the messages fail to merge, panic?
+        // write the events to the db
+        // Commit the transaction
+        // Emit events
+    }
+
+    fn commit_block(&mut self, block: Block) {
+        // Commit the individual shards state changes
+        for shard_chunks in block.shard_chunks.clone() {
```
suurkivi, line 94: will this be an expensive clone()? (not saying it's necessarily bad, just curious)

src/consensus/validator.rs
sanjayprabhu: Fixed, but also removed the error. If this fails, there's no recovering, consensus is going to halt, and the error is not handled anyway, so might as well panic.

src/storage/store/engine.rs
sanjayprabhu: Yes, this is probably expensive. We're using clones too liberally in general. We shouldn't be cloning the blocks/chunks/transactions. Depending on blocksize, it could be very costly. We need to operate on references.

Will fix in followups.

src/storage/store/engine.rs
```diff
@@ -0,0 +1,121 @@
...
+use tracing::error;
+
+// Shard state root and the transactions
+pub struct ShardStateChange {
+    pub shard_id: u32,
+    pub new_state_root: Vec<u8>,
+    pub transactions: Vec<proto::Transaction>,
+}
+
+pub trait Engine {
```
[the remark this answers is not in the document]
sanjayprabhu: No, I was thinking of proposer/validator tests which will need to mock out the engine.

PR #37, opened 2024-11-15
separate shard and block stores and write to shard store

aditiharini:
Initialize a db per-shard and write shard chunks separately from blocks.

### DIFF
--- a/src/consensus/proposer.rs
+++ b/src/consensus/proposer.rs
@@ -5,7 +5,7 @@ use crate::core::types::{
 use crate::proto::rpc::snapchain_service_client::SnapchainServiceClient;
 use crate::proto::rpc::BlocksRequest;
 use crate::proto::snapchain::{Block, BlockHeader, FullProposal, ShardChunk, ShardHeader};
-use crate::storage::store::engine::{Engine, ShardStateChange, SnapchainEngine};
+use crate::storage::store::engine::{ShardEngine, ShardStateChange, SnapchainEngine};
 use crate::storage::store::BlockStorageError;
 use malachite_common::{Round, Validity};
 use prost::Message;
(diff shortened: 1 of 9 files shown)


src/storage/store/engine.rs
```diff
@@ -89,33 +81,52 @@ impl Engine for SnapchainEngine {
...
+impl SnapchainEngine {
+    pub fn new(block_shard: u32, block_store: BlockStore) -> Self {
+        SnapchainEngine {
+            block_shard,
+            block_store,
         }
+    }
+
+    pub fn commit_block(&mut self, block: Block) {
+        // Commit the individual shards state changes
```
aditiharini, line 104: Do we want to commit shards when the block gets committed?

src/storage/store/utils.rs
```diff
@@ -1,3 +1,4 @@
+pub const PAGE_SIZE_MAX: usize = 10_000;
```
sanjayprabhu, line 1: This is fine for messages, but way too high for blocks and shards, should be more like 100-1000 for them

src/storage/store/engine.rs
```diff
@@ -80,42 +74,53 @@ impl Engine for SnapchainEngine {
...
+    }
+}
+
+pub struct BlockEngine {
+    block_store: BlockStore,
+    shard_id: u32,
+}
+
+impl BlockEngine {
+    pub fn new(shard_id: u32, block_store: BlockStore) -> Self {
```
sanjayprabhu, line 106: Since we know this is always for the block, we don't need to accept the shard id, can use use a constant of 0.

src/consensus/proposer.rs
```diff
@@ -53,14 +53,14 @@ pub struct ShardProposer {
     chunks: Vec<ShardChunk>,
     proposed_chunks: BTreeMap<ShardHash, FullProposal>,
     tx_decision: Option<TxDecision>,
-    engine: SnapchainEngine, // TODO: Use the trait here, once we figure out how to pass Box<dyn Engine> to the actor
+    engine: ShardEngine, // TODO: Use the trait here, once we figure out how to pass Box<dyn Engine> to the actor
```
sanjayprabhu, line 56: Can remove this comment

sanjayprabhu (approved):

PR #43, opened 2024-11-18
fully integrate block and shard stores

aditiharini:
Remove the in memory versions of the block and shard stores and rely entirely on the db.


src/consensus/proposer.rs
```diff
@@ -230,15 +267,14 @@ impl BlockProposer {
...
-                       if let Some(proto::full_proposal::ProposedValue::Shard(chunk)) = decision.proposed_value {
-                            let chunk_height = chunk.header.clone().unwrap().height.unwrap();
-                            let chunk_block_number = chunk_height.block_number;
-                            if self.pending_chunks.contains_key(&chunk_block_number) {
-                                self.pending_chunks.get_mut(&chunk_block_number).unwrap().push(chunk);
-                            } else {
-                                self.pending_chunks.insert(chunk_block_number, vec![chunk]);
-                            }
+                    // TODO(aditi): This breaks if syncd shard chunks show up in shard_decision_rx.
+                    if let Ok(chunk) = self.shard_decision_rx.try_recv() {
```
sanjayprabhu, line 271: This makes me think we should separate sync and consensus channels. We might want to start up consensus, buffer all the messages while sync completes and then process them. No need to change anything right now, but something we'll need to consider how to handle.

src/consensus/proposer.rs
```diff
@@ -158,6 +168,35 @@ impl Proposer for ShardProposer {
...
+                    let mut rpc_client = SnapchainServiceClient::connect(destination_addr).await?;
+                    let request = Request::new(ShardChunksRequest {
+                        shard_id: self.shard_id.shard_id(),
+                        start_block_number: prev_block_number + 1,
+                        stop_block_number: None,
+                    });
+                    let missing_shard_chunks = rpc_client.get_shard_chunks(request).await?;
+                    for shard_chunk in missing_shard_chunks.get_ref().shard_chunks.clone() {
+                        self.engine.commit_shard_chunk(shard_chunk.clone());
+                        self.publish_new_shard_chunk(shard_chunk).await;
```
sanjayprabhu, line 192: I think we shouldn't publish here, or at leas use a separate channel that consensus is not listening to.

src/node/snapchain_node.rs
```diff
@@ -43,10 +44,12 @@ impl SnapchainNode {
 
         let mut consensus_actors = BTreeMap::new();
 
-        let (shard_decision_tx, shard_decision_rx) = mpsc::channel::<Decision>(100);
+        let (shard_decision_tx, shard_decision_rx) = mpsc::channel::<ShardChunk>(100);
 
         let mut shard_messages: HashMap<u32, mpsc::Sender<message::Message>> = HashMap::new();
 
+        let mut shard_stores = HashMap::new();
```
suurkivi, line 51: Still getting used to rust, but I would have a preference here to annotate the type of this for readability.

- it is being used in a fairly large function with some uses that are far away
- it's not immediately clear what will go in there

I guess my editor can figure it out if I hover over it, but above comments still hold

src/storage/store/engine.rs
```diff
@@ -113,6 +123,17 @@ impl BlockEngine {
         }
     }
 
+    pub fn get_last_block(&self) -> Option<Block> {
```
suurkivi, line 126: should this return a result and push the decision to log, etc, to the caller?

src/consensus/proposer.rs
```diff
@@ -43,14 +43,19 @@ pub trait Proposer {
...
+    ) -> Result<(), Box<dyn std::error::Error>>;
 }
 
 pub struct ShardProposer {
     shard_id: SnapchainShard,
     address: Address,
     chunks: Vec<ShardChunk>,
     proposed_chunks: BTreeMap<ShardHash, FullProposal>,
-    tx_decision: Option<TxDecision>,
+    tx_decision: Option<mpsc::Sender<ShardChunk>>,
```
suurkivi, line 58: can we remove Option here?

src/storage/store/shard.rs
```diff
@@ -118,25 +124,70 @@ pub fn put_shard_chunk(db: &RocksDB, shard_chunk: ShardChunk) -> Result<(), Shar
...
+
     pub fn max_block_number(&self) -> Result<u64, ShardStorageError> {
         let current_height = get_current_height(&self.db)?;
         match current_height {
             None => Ok(0),
             Some(height) => Ok(height),
         }
     }
+
+    pub fn get_shard_chunks(
```
suurkivi, line 165: maybe it's time for some unit tests on things like this? Then again maybe it's too early. Perhaps some TODOs where we'd like to have unit tests (assuming these functions survive further refactoring).

src/storage/store/engine.rs
aditiharini: Can we make this change after we decide our error-handling type situation. If we want to use HubError, I'll change to that and expose HubError in all our interfaces.

src/storage/store/shard.rs
aditiharini: Yeah agreed-- I think it's worth unit testing. I'll write tests after we discuss how to handle the sync edge cases today? That might change this interface somewhat.

PR #44, opened 2024-11-18
First engine tests

suurkivi:


src/storage/store/engine_tests.rs
```diff
@@ -0,0 +1,249 @@
...
+    fn test_engine_basic_propose() {
+        let mut engine = new_engine();
+        let state_change = engine.propose_state_change(1);
+
+        assert_eq!(1, state_change.shard_id);
+        assert_eq!(state_change.transactions.len(), 1);
+        assert_eq!(0, state_change.transactions[0].user_messages.len());
+        assert_eq!(
+            "237b11d0dd9e78994ef2f141c7f170d48bb51d34",
+            to_hex(&state_change.new_state_root)
```
sanjayprabhu, line 114: Hardcoding hashes is going to be fragile, especially since everything is in flux. I think we should just check that the root hash in the merkle trie matches the hash in the state change struct and not care about the actual values.

suurkivi: Yes, agree with all above.

But I actually want this to be quite fragile to start, if the values change for any reason it will be interesting to see why they're changing and will require manual investigation and updates.

I'm happy to leave a comment to future editors of the test to feel free to do what you suggest above becomes too tedious and we have confidence with the code.

src/storage/store/engine_tests.rs
```diff
@@ -0,0 +1,249 @@
...
+    use ed25519_dalek::{SecretKey, SigningKey};
+    use hex::FromHex;
+    use tracing_subscriber::EnvFilter;
+
+    fn new_engine() -> ShardEngine {
+        let dir = tempfile::TempDir::new().unwrap();
+        let db_path = dir.path().join("a.db");
+
+        let db = db::RocksDB::new(db_path.to_str().unwrap());
+        db.open().unwrap();
```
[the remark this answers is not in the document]
suurkivi: Yes I believe so but actually you uncovered a bug.

TempDir is deleting files out from under the db as soon as this function returns. I need to keep it in scope. Fixing.

PR #46, opened 2024-11-19
Store refactor

sanjayprabhu:
Move around some functions and constants and fix compile errors in the cast store so we can start integrating it into the engine


src/storage/store/block.rs
```diff
@@ -66,7 +61,8 @@ fn get_block_page_by_prefix(
         }
 
         Ok(false) // Continue iterating
-    })?;
+    })
+    .map_err(|e| BlockStorageError::TooManyBlocksInResult)?; // TODO: Return the right error
```
sanjayprabhu, line 65: We need to think about how to handle errors. Existing hub code is using HubError everywhere. This is causing some pain when we integrate. Maybe we should stick with HubError everywhere for now and then revisit later? This HubError is used all the way to the clients, so if we switch to something else, we still need to map to a HubError for backwards compatibility.

src/storage/constants.rs
```diff
@@ -0,0 +1,94 @@
+pub const PAGE_SIZE_MAX: usize = 1_000;
+
+/** Copied from the JS code */
```
suurkivi, line 3: perhaps this comment is no longer quite correct

src/storage/constants.rs
```diff
@@ -0,0 +1,94 @@
...
+#[allow(dead_code)]
+pub enum RootPrefix {
+    Block = 1,
```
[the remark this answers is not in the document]
sanjayprabhu: No benefit I think, in the reconciliation case, we'd compare message hashes, not db key/values directly.

PR #47, opened 2024-11-19
External transaction experiment

suurkivi:


src/storage/trie/merkle_trie.rs
```diff
@@ -159,18 +169,32 @@ impl MerkleTrie {
      *  Note: We require a write-locked root node to perform this operation, which should
      *  be supplied by the caller.
      */
-    fn unload_from_memory(&self, root: &mut TrieNode, force: bool) -> Result<(), HubError> {
```
suurkivi, line 162: just doing some refactoring here to remove this force: flag since it was always `true`

src/storage/store/engine.rs
```diff
@@ -167,7 +171,15 @@ impl ShardEngine {
         );
 
         if hashes_match {
-            self.trie.commit().unwrap();
+            {
+                let mut write_guard = self.db.db.write().unwrap();
+                if let Some(db) = write_guard.as_mut() {
+                    let mut txn = db.transaction();
+                    self.trie.commit_to_txn(&mut txn);
+                    txn.commit().unwrap();
```
sanjayprabhu, line 179: I think the control should be the other way around, we provide it the RocksDBTransaction batch and we commit it outside the trie. The stores also operate on the RocksDBTransaction batch

src/storage/trie/merkle_trie.rs
```diff
@@ -128,7 +129,16 @@ impl MerkleTrie {
 
     pub fn commit(&self) -> Result<(), HubError> {
         if let Some(root) = self.root.write().unwrap().as_mut() {
-            self.unload_from_memory(root, true)
+            self.unload_from_memory(root)
```
sanjayprabhu, line 132: The DB commit should happen outside the trie, and maybe the trie should just always unload from memory?

This is going to cause performance issues, we'll likely need some kind of LRU cache for frequently accessed fids. But we can worry about that later.

src/storage/store/engine.rs
```diff
@@ -153,7 +157,13 @@ impl ShardEngine {
             "commit insert"
         );
 
-        self.trie.insert(hashes.clone()).unwrap();
+        let db = &*self.shard_store.db;
```
sanjayprabhu, line 160: We should get the db from self instead of reaching into shard_store. Might be a good idea to ensure no one except the engine needs to hold a reference to the db in their struct. But don't need to fix it in this pr.

PR #48, opened 2024-11-19
sync on first proposal

aditiharini:
Sync blocks on the first proposal a node sees. This is to catch cases where blocks are missed between the initial sync on `RegisterValidator` and the first proposal.


src/consensus/consensus.rs
```diff
@@ -327,7 +327,7 @@ impl Consensus {
                 // println!("Connected to {connected_peers}/{total_peers} peers");
                 state
                     .shard_validator
-                    .sync_with_new_validator(&validator)
+                    .sync_against_validator(&validator)
```
sanjayprabhu, line 330: I think we should move this outside consensus.

Can we delete this for now? Or do we need it both in register and in ReceivedFullProposal?

PR #50, opened 2024-11-19
validate first version

suurkivi:


src/storage/store/engine.rs
```diff
@@ -162,88 +182,99 @@ impl ShardEngine {
...
-                            );
-                        }
-                        Err(err) => {
-                            error!("Unable to merge cast message during commit: {}", err);
-                            // If we're unable to merge during a commmit, it's going to cause a consensus halt. This state change should've been validated before.
-                            panic!("Unable to merge cast message during commit");
-                        }
-                    }
+                    cast_store.merge(msg, txn_batch).map_err(|err| {
+                        EngineError::MergeCastMessageError(format!(
```
sanjayprabhu, line 199: This should just be a MergeMessageError, but we can refactor later.

sanjayprabhu (approved):
Nice

PR #54, opened 2024-11-20
Update collect_stats.rs

savvar9991:
Added more descriptive comments to enhance the understanding of the logic, making it easier for future maintainers.
 Ensured that variable names are more self-explanatory (e.g., changed num_messages_submitted and num_messages_confirmed to more precise names based on context).
Ensured that indentation and spacing are consistent to improve readability. Removed unnecessary .clone() calls where the values could be borrowed or handled without duplication, reducing memory overhead.
 The submit_message_timer is simplified, and the timer now only ticks once per interval, instead of repeatedly checking for an interval every loop iteration.
Instead of creating and clearing vectors for each stats calculation interval, I ensured that the stats collection logic is more efficient by handling them in batch operations.
 I added Result handling for asynchronous operations that might fail, such as in SnapchainServiceClient::connect. This helps prevent unhandled panics and ensures that failures are managed more gracefully.
Added a more efficient way to handle the termination of tasks and ensure that all background tasks (submit_message_handles, follow_blocks_handle) are aborted when the scenario finishes. The use of select! is already good for managing multiple asynchronous streams, but I ensured that tasks are handled as efficiently as possible, reducing overhead. Improved the logging in the stats_calculation_timer.tick block to give more comprehensive output on the benchmarking statistics. Added additional context to the print statements for better insights into how the benchmarking is performing, especially around message submission, block production, and confirmation times.

### DIFF
--- a/src/bin/collect_stats.rs
+++ b/src/bin/collect_stats.rs
@@ -11,8 +11,9 @@ use snapchain::utils::cli::{compose_message, follow_blocks, send_message};
 use std::collections::HashSet;
 use std::time::Duration;
 use tokio::sync::mpsc;
-use tokio::time::Instant;
+use tokio::time::{Instant, interval};
 use tokio::{select, time};
+
 const STATS_CALCULATION_INTERVAL: Duration = Duration::from_secs(10);
 const NUM_ITERATIONS: u64 = 5;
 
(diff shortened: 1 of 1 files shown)


suurkivi:
Not accepting external PRs at this time

PR #55, opened 2024-11-20
Update proposer.rs

savvar9991:
In many cases, cloning values when they aren't necessary was redundant. I've removed unnecessary cloning and passed references directly when possible. 
Improved error logging, especially for potential network-related failures. The previous error handling didn't capture the specific RPC errors effectively.
 Used tracing::warn! and tracing::error! more consistently for error and warning logs. 
 Used more explicit handling of timeouts and the polling of shard chunks, ensuring that the timeout logic doesn't end up hanging forever. 
 Simplified the timeout logic by using select! to handle poll_interval and timeout more efficiently. 
 Replaced unwrap calls with more graceful error handling, ensuring that the code doesn't panic unexpectedly Streamlined function signatures and used destructuring where appropriate to make the code easier to follow. 
 Grouped similar operations together (e.g., hashing logic, proposal creation) for better clarity.

### DIFF
--- a/src/consensus/proposer.rs
+++ b/src/consensus/proposer.rs
@@ -1,6 +1,4 @@
-use crate::core::types::{
-    proto, Address, Height, ShardHash, ShardId, SnapchainShard, SnapchainValidator,
-};
+use crate::core::types::{proto, Address, Height, ShardHash, SnapchainShard, SnapchainValidator};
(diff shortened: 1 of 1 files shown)


suurkivi:
Not accepting external PRs at this time

PR #58, opened 2024-11-20
Update collect_stats.rs

savvar9991:
1) Fixed num_messages_submitted tracking: The number of submitted messages is now tracked in the submit_messages function and passed into calculate_stats.
2) Fix for stopping after NUM_ITERATIONS: The simulation will now stop after NUM_ITERATIONS messages have been submitted.
3) Corrected blocks_rx.clone() issue: You can’t clone a Receiver, so you should rethink the logic around receiving messages. You should use Receiver in a single task at a time.
4) Graceful error handling: Using map_err to return errors instead of panicking with unwrap().
5) Added select! import: tokio::select is now properly imported for asynchronous waiting.

### DIFF
--- a/src/bin/collect_stats.rs
+++ b/src/bin/collect_stats.rs
@@ -1,6 +1,5 @@
 use ed25519_dalek::{SecretKey, SigningKey};
-use hex;
-use hex::FromHex;
+use hex::{self, FromHex};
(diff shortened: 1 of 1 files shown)


suurkivi:
looks automated

PR #60, opened 2024-11-20
Update proposer.rs

savvar9991:
1)The method now returns a Result type, ensuring that any errors encountered during block publication (e.g., network or internal processing errors) are returned instead of just being logged. If publishing the block fails, it returns a BlockPublishError. 2)When connecting to the validator's RPC endpoint, connection errors are now captured and returned as a RpcTransportError. Response errors from the RPC server (such as a malformed response) are caught and wrapped as RpcResponseError. These changes improve the error propagation by returning errors instead of continuing silently or simply logging them.


suurkivi:
looks automated

PR #63, opened 2024-11-20
Update setup.rs

savvar9991:
-Removed excessive use of format! in places where it is not necessary. 
-Added error handling for file writing.
-Converted the secret key to bytes before encoding it in hex.
-Added the missing propose_value_delay variable to the configuration string.


suurkivi:
looks automated

PR #66, opened 2024-11-21
store on chain events from the engine

aditiharini:
Start storing onchain events. Onchain events still aren't plumbed through consensus, but we can manually feed them to the engine and test behavior.

### DIFF
--- a/src/storage/store/account/onchain_event_store.rs
+++ b/src/storage/store/account/onchain_event_store.rs
@@ -1,0 +1,40 @@
+use std::sync::Arc;
+
+use prost::Message;
+
+use super::make_fid_key;
+use crate::proto::snapchain::OnChainEvent;
+use crate::storage::db::{RocksDB, RocksdbError};
+use thiserror::Error;
+
+#[derive(Error, Debug)]
+pub enum OnchainEventStorageError {
+    #[error(transparent)]
+    RocksdbError(#[from] RocksdbError),
+}
+
+pub fn put_onchain_event(
+    db: &RocksDB,
+    onchain_event: OnChainEvent,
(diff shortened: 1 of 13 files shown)


src/storage/store/account/onchain_event_store.rs
```diff
@@ -0,0 +1,63 @@
...
+    #[error(transparent)]
+    HubError(#[from] HubError),
+}
+
+pub fn merge_onchain_event(
+    db: &RocksDB,
+    onchain_event: OnChainEvent,
+) -> Result<(), OnchainEventStorageError> {
+    let primary_key = make_fid_key(onchain_event.fid as u32);
+    db.put(&primary_key, &onchain_event.encode_to_vec())?;
```
sanjayprabhu, line 27: This needs the OnChainEvent RootPrefix, right?

src/storage/store/engine.rs
```diff
@@ -45,9 +50,12 @@ pub struct ShardEngine {
     shard_store: ShardStore,
     messages_rx: mpsc::Receiver<message::Message>,
     messages_tx: mpsc::Sender<message::Message>,
+    validator_message_rx: mpsc::Receiver<snapchain::ValidatorMessage>,
+    validator_message_tx: mpsc::Sender<snapchain::ValidatorMessage>,
```
sanjayprabhu, line 54: I think we should re-use the existing channel. Change the type to be an enum that can hold a Message or a ValidatorMessage.

src/storage/store/account/onchain_event_store.rs
aditiharini: Copied the same primary key prefix we use in hub now.

PR #68, opened 2024-11-21
Support deletes

sanjayprabhu:
Reflect store operations in the trie and refactor engine to create multiple transactions.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -47,6 +47,7 @@ threadpool = "1.8.1"
 rand = "0.8.5"
 humantime-serde = "1.1.1"
 humantime = "2.1.0"
+itertools = "0.13.0"
 
 [build-dependencies]
 tonic-build = "0.9.2"
--- a/src/core/message.rs
+++ b/src/core/message.rs
@@ -29,14 +29,12 @@ impl message::Message {
 
 impl snapchain::ValidatorMessage {
     pub fn fid(&self) -> u32 {
-        if self.fname_transfer.is_some() {
-            return self.fname_transfer.as_ref().unwrap().fid as u32;
+        if let Some(fname) = &self.fname_transfer {
+            return fname.fid as u32;
         }
-
-        if self.on_chain_event.is_some() {
-            return self.on_chain_event.as_ref().unwrap().fid as u32;
+        if let Some(event) = &self.on_chain_event {
+            return event.fid as u32;
         }
-
         0
     }
 }
--- a/src/core/message.rs
+++ b/src/core/message.rs
@@ -1,4 +1,5 @@
 use crate::proto::msg as message;
+use crate::proto::snapchain;
 
 impl message::Message {
     pub fn is_type(&self, message_type: message::MessageType) -> bool {
@@ -19,5 +20,23 @@ impl message::Message {
         } else {
             0
         }
+    }
+
+    pub fn hex_hash(&self) -> String {
+        hex::encode(&self.hash)
+    }
+}
+
+impl snapchain::ValidatorMessage {
+    pub fn fid(&self) -> u32 {
+        if self.fname_transfer.is_some() {
+            return self.fname_transfer.as_ref().unwrap().fid as u32;
+        }
+
+        if self.on_chain_event.is_some() {
+            return self.on_chain_event.as_ref().unwrap().fid as u32;
+        }
+
+        0
     }
 }
--- a/src/storage/store/engine.rs
+++ b/src/storage/store/engine.rs
@@ -1,10 +1,9 @@
-use super::account::{OnchainEventStorageError, OnchainEventsPage};
+use super::account::OnchainEventStorageError;
 use super::shard::ShardStore;
 use crate::core::error::HubError;
-use crate::core::types::{proto, Height};
+use crate::core::types::Height;
 use crate::proto::onchain_event::{OnChainEvent, OnChainEventType};
 use crate::proto::{hub_event, msg as message, snapchain};
-use crate::storage::db;
 use crate::storage::db::{PageOptions, RocksDB, RocksDbTransactionBatch};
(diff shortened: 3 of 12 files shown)


src/storage/store/engine.rs
```diff
@@ -308,6 +390,15 @@ impl ShardEngine {
         }
     }
 
+    pub fn sync_id_exists(&mut self, sync_id: &Vec<u8>) -> bool {
```
suurkivi, line 393: Is this for test use, or public use? If test, I'd go with pub(crate).

Otherwise, probably best to surface the error back to the caller instead of logging and returning false. I think trie.exists should only fail under extreme conditions.

Cargo.toml
```diff
@@ -47,6 +47,7 @@ threadpool = "1.8.1"
 rand = "0.8.5"
 humantime-serde = "1.1.1"
 humantime = "2.1.0"
+itertools = "0.13.0"
```
suurkivi, line 50: don't think this made it into the PR?

sanjayprabhu: It did. This is what implements into_group_map_by

src/storage/store/engine.rs
sanjayprabhu: Test, I'll fix

src/core/message.rs
```diff
@@ -20,4 +21,22 @@ impl message::Message {
...
         }
     }
+
+    pub fn hex_hash(&self) -> String {
+        hex::encode(&self.hash)
+    }
+}
+
+impl snapchain::ValidatorMessage {
+    pub fn fid(&self) -> u32 {
```
suurkivi, line 31: Two things:

1) I might prefer something like below for improved readability:

```rust
impl snapchain::ValidatorMessage {
    pub fn fid(&self) -> u32 {
        if let Some(fname) = &self.fname_transfer {
            return fname.fid as u32;
        }
        if let Some(event) = &self.on_chain_event {
            return event.fid as u32;
        }
        0
    }
}
```

I think seeing the unwrap there I really have to scrutinize the code, and it took me a moment to realize that it was indeed safe due to the is_some() above.

I think we can probably remove most of the unwraps() in this file with similar refactors and it will be MUCH faster to reason about.

2) What to do in the fall through case? Should we return an error or even panic? We don't expect this to ever be either? Can we model as actual enum in protobuf?

Cargo.toml
suurkivi: Oh, it's one of those things that adds magic behind the scenes. Still getting used to that.

src/storage/store/engine.rs
```diff
@@ -145,75 +156,65 @@ impl ShardEngine {
...
-                        }
-                        unhandled_type => {
-                            return Err(EngineError::UnsupportedMessageType(unhandled_type));
-                        }
-                    }
-                }
-            }
+        let mut snap_txns = self.create_transactions_from_mempool(messages);
+        for snap_txn in &mut snap_txns {
+            self.replay_snap_txn(&snap_txn, txn_batch)?;
```
suurkivi, line 161: ok, this is really nice, same logic applied in all three of the main functions.

src/storage/store/engine.rs
```diff
@@ -145,75 +156,65 @@ impl ShardEngine {
...
-                                .insert(&*self.db, txn_batch, vec![msg.hash.clone()])?;
-                        }
-                        unhandled_type => {
-                            return Err(EngineError::UnsupportedMessageType(unhandled_type));
-                        }
-                    }
-                }
-            }
+        let mut snap_txns = self.create_transactions_from_mempool(messages);
+        for snap_txn in &mut snap_txns {
```
suurkivi, line 160: nit: is snap_txn short for snapchain_txn? I might prefer the longer form as snap_txn implies (on first read) to be introducing a new concept.

src/storage/store/engine.rs
```diff
@@ -224,48 +225,129 @@ impl ShardEngine {
...
+    }
+
+    fn update_trie(
+        &mut self,
+        event: hub_event::HubEvent,
+        txn_batch: &mut RocksDbTransactionBatch,
+    ) -> Result<(), EngineError> {
+        match event.body {
+            Some(hub_event::hub_event::Body::MergeMessageBody(merge)) => {
+                if let Some(msg) = merge.message {
```
suurkivi, line 329: Do we expect merge.message to ever be None here? Should we return an error or panic in the missing case, or do you think it's ok or even normal to ignore?

src/core/message.rs
sanjayprabhu: 1. Ah, much cleaner. Fixed.
2. I think if we use oneof it will use enums. We should probably switch to that actually. But even if we did, it won't solve the problem because it's acceptable to have 0 as a value in the protobuf, so callers have to handle fid 0 (which is not valid for the application, fids start from 1). So, I think defaulting to 0 is fine here.

src/storage/store/engine.rs
sanjayprabhu: No, it's possible for it to be None. e.g. registering your fid for the first time will generate only an onchain event, or similarly, registering your fname will only have an fname transfer.

src/storage/store/engine.rs
```diff
@@ -224,48 +225,129 @@ impl ShardEngine {
...
+                if let Some(onchain_event) = merge.on_chain_event {
+                    self.trie.insert(
+                        &self.db,
+                        txn_batch,
+                        vec![onchain_event.transaction_hash.clone()],
+                    )?;
+                }
+            }
+            _ => {
+                return Err(EngineError::UnsupportedEvent);
```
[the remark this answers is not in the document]
sanjayprabhu: This is a placeholder until we implement the other event types, we will not have the fallback case at all. So not worth it. I'll add a todo.

PR #81, opened 2024-11-25
implement event stream channel and basic rpc

aditiharini:
Added support for an event stream rpc that emits new events and added unit tests.


src/storage/store/engine.rs
```diff
@@ -382,19 +394,30 @@ impl ShardEngine {
...
+                    match self.events_tx.send(event) {
+                        Ok(_) => {}
+                        Err(err) => {
+                            error!("Unable to broadcast event {:#?}", err)
+                        }
+                    }
+                }
+                self.trie.reload(&self.db).unwrap();
+
+                match self.shard_store.put_shard_chunk(shard_chunk) {
```
sanjayprabhu, line 410: We should pull all this to it's own function like `commit_and_emit_events`

src/storage/store/engine.rs
```diff
@@ -269,7 +280,8 @@ impl ShardEngine {
 
                 match event {
                     Ok(hub_event) => {
-                        self.update_trie(hub_event, txn_batch)?;
+                        self.update_trie(hub_event.clone(), txn_batch)?;
+                        events.push(hub_event.clone());
```
sanjayprabhu, line 284: We're cloning three times in one function, can we avoid the clone by using references everywhere?

src/storage/store/account/event.rs
```diff
@@ -143,7 +143,7 @@ impl HubEvent {
         let start_prefix = Self::make_event_key(start_id);
         let stop_prefix = match stop_id {
             Some(id) => Self::make_event_key(id),
-            None => increment_vec_u8(&vec![RootPrefix::HubEvents as u8 + 1]),
```
aditiharini, line 146: We probably have this issue in other stores and need tests.

PR #82, opened 2024-11-25
add statsd metrics

suurkivi:


src/bin/setup_local_testnet.rs
```diff
@@ -42,23 +49,30 @@ async fn main() {
...
+
         let config_file_content = format!(
             r#"
 id = {id}
 rpc_address="{rpc_address}"
 rocksdb_dir="{db_dir}"
 
+[metrics]
+prefix="{metrics_prefix}"
+addr="{metrics_addr}"
```
sanjayprabhu, line 74: We should call this statsd_host/statsd_addr

src/storage/store/engine.rs
```diff
@@ -203,13 +210,30 @@ impl ShardEngine {
...
+
+    fn count(&self, key: &str, value: u64) {
+        let key = format!("shard{}.{}", self.shard_id, key);
+        _ = self.metrics_client.count(key.as_str(), value)
+    }
+
+    fn gauge(&self, key: &str, value: u64) {
+        let key = format!("shard{}.{}", self.shard_id, key);
+        _ = self.metrics_client.gauge(key.as_str(), value)
+    }
```
sanjayprabhu, line 226: I think we should do this in a wrapper around the metrics client (similar to logging spans). 

Also, I think the shard indexes should be tags.

PR #91, opened 2024-11-26
migrate link store

aditiharini:
Support link messages in the engine.

### DIFF
--- a/src/storage/store/account/link_store.rs
+++ b/src/storage/store/account/link_store.rs
@@ -148,18 +148,22 @@ impl LinkStore {
         r#type: String,
         page_options: &PageOptions,
     ) -> Result<MessagesPage, HubError> {
-        let prefix: Vec<u8> = LinkStore::links_by_target_key(target, 0, None)?;
+        let start_prefix: Vec<u8> = LinkStore::links_by_target_key(target, 0, None)?;
+        let stop_target = match target {
+            Target::TargetFid(fid) => Target::TargetFid(fid + 1),
+        };
+        let stop_prefix = LinkStore::links_by_target_key(&stop_target, 0, None)?;
 
         let mut message_keys = vec![];
         let mut last_key = vec![];
 
         store.db().for_each_iterator_by_prefix(
-            Some(prefix.to_vec()),
-            Some(increment_vec_u8(&prefix)),
+            Some(start_prefix.to_vec()),
+            Some(stop_prefix.to_vec()),
             page_options,
             |key, value| {
                 if r#type.is_empty() || value.eq(r#type.as_bytes()) {
-                    let ts_hash_offset = prefix.len();
+                    let ts_hash_offset = start_prefix.len();
                     let fid_offset: usize = ts_hash_offset + TS_HASH_LENGTH;
 
                     let fid =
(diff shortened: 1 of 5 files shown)


src/storage/store/account/link_store.rs
```diff
@@ -151,51 +127,52 @@ impl LinkStore {
...
-        store: &Store,
+        store: &Store<LinkStore>,
         target: &Target,
         r#type: String,
         page_options: &PageOptions,
     ) -> Result<MessagesPage, HubError> {
-        let prefix: Vec<u8> = LinkStore::links_by_target_key(target, 0, None)?;
+        let start_prefix: Vec<u8> = LinkStore::links_by_target_key(target, 0, None)?;
+        let stop_target = match target {
+            Target::TargetFid(fid) => Target::TargetFid(fid + 1),
```
sanjayprabhu, line 153: Use increment_vec_u8, which is more generic.

src/storage/store/account/link_store.rs
```diff
@@ -205,12 +182,12 @@ impl LinkStore {
...
                 Ok(false)
-            })?;
+            },
+        )?;
 
-        let messages_bytes =
-            message::get_many_messages_as_bytes(store.db().borrow(), message_keys)?;
+        let messages_bytes = get_many_messages_as_bytes(store.db().borrow(), message_keys)?;
         let next_page_token = if last_key.len() > 0 {
-            Some(last_key[prefix.len()..].to_vec())
```
sanjayprabhu, line 213: Hmm, why did this change?

src/storage/store/account/link_store.rs
```diff
@@ -502,46 +479,8 @@ impl LinkStore {
...
-
-            deferred_settle_messages(deferred, &channel, messages);
-        });
-
-        Ok(promise)
-    }
-
     pub fn get_link_removes_by_fid(
-        store: &Store,
+        store: &Store<LinkStore>,
```
sanjayprabhu, line 483: Why did these change? Not a big deal, but we should keep it consistent with hubs if possible. The CastStore uses just Store

aditiharini: I changed the cast store too earlier. The reason i did this is because deriving works better with the generic. I wanted to derive clone and couldn't do it with the version that existed before because of the way we embedded the store. with `Box` and `dyn`.

src/storage/store/account/link_store.rs
aditiharini: This changed at the db level when I pulled the db code in. The page token is the whole key and not a piece of it. Only providing the suffix in the token worked when we only iterated over a single prefix (e.g. fid) but now that we have a start and stop we can't keep only some suffix.

src/storage/store/account/link_store.rs
aditiharini: I actually preferred this method. With `increment_vec_u8` you need to be sure that the key you're incrementing is the last part of the prefix otherwise it won't do what you want. Here, I actually had `increment_vec_u8` first but then checked the key structure and saw that the target fid is somewhere in the middle of the prefix so I changed it.

aditiharini: I actually just made the change you suggested in my next feature. In that one, there's some string field we index on and it was hard to figure out how to increment that. I just changed this to be consistent too.

PR #96, opened 2024-11-27
fix/silence all warnings

suurkivi:


sanjayprabhu:
Amazing, thank you. We should have CI  fail if there are warnings.

PR #111, opened 2024-12-02
migrate username proof store

suurkivi:


src/storage/store/engine_tests.rs
```diff
@@ -541,6 +542,47 @@ mod tests {
...
+        }
+        {
+            let username_proof_result = engine.get_username_proofs_by_fid(FID_FOR_TEST);
+            assert!(username_proof_result.is_ok());
+
+            let messages_bytes_len = username_proof_result.unwrap().messages_bytes.len();
+            assert_eq!(1, messages_bytes_len);
+        }
+
+        // TODO: test get_username_proof (by name)
```
suurkivi, line 582: do we need these two things? if so, now?

src/storage/store/engine_tests.rs
```diff
@@ -541,6 +542,47 @@ mod tests {
...
+        // TODO: do we need a test for ENS name registration?
```
sanjayprabhu, line 583: We can do this when we do the engine level validation for ens names

src/storage/store/engine_tests.rs
sanjayprabhu: No, we can skip for now. I think we might need it for RPC api compatibility, but we can add it then.

sanjayprabhu (approved):

PR #161, opened 2024-12-13
More read apis

sanjayprabhu:
Implement read apis for most of the remaining stores

### DIFF
--- a/src/network/server_tests.rs
+++ b/src/network/server_tests.rs
@@ -226,7 +226,7 @@ mod tests {
         let cast_remove = messages_factory::casts::create_cast_remove(
             SHARD1_FID,
             &cast_add.hash,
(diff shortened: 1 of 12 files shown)


src/network/server_tests.rs
```diff
@@ -276,5 +276,58 @@ mod tests {
...
+            page_size: Some(1),
+            page_token: None,
+            reverse: None,
+        };
+        let response = service
+            .get_casts_by_fid(Request::new(all_casts_request))
+            .await;
+        test_helper::assert_contains_all_messages(&response, &[&cast_add2]);
+
+        // TODO: Fix pagination
```
sanjayprabhu, line 304: I'm not entirely sure why but I think the rocksdb pagination is broken. Will investigate tomorrow. cc @aditiharini

PR #164, opened 2024-12-14
make store type names consistent with hubs

aditiharini:
The `StoreType` names in hubs don't have the `STORE_TYPE` prefix. Manually translate the enum to strings to match hubs.


src/storage/store/stores.rs
```diff
@@ -248,9 +248,18 @@ impl Stores {
             let max_messages =
                 self.store_limits
                     .max_messages(slot.units, slot.legacy_units, store_type);
+            let name = match store_type {
```
sanjayprabhu, line 251: Clean to do a string replace instead of enumerating manually. But not a big deal.

PR #169, opened 2024-12-17
Handle conflicting messages within the same transaction

sanjayprabhu:
Because we only commit to the db all at once at the end of the commit phase, we were not correctly handling conflicting messages within the same transaction. 

This looks in the transaction batch to find conflicting keys before falling back to the db.

### DIFF
--- a/src/storage/store/account/message.rs
+++ b/src/storage/store/account/message.rs
@@ -156,14 +156,35 @@ pub fn make_cast_id_key(cast_id: &CastId) -> Vec<u8> {
 
 pub fn get_message(
     db: &RocksDB,
+    txn: &mut RocksDbTransactionBatch,
     fid: u64,
     set: u8,
     ts_hash: &[u8; TS_HASH_LENGTH],
 ) -> Result<Option<MessageProto>, HubError> {
     let key = make_message_primary_key(fid, set, Some(ts_hash));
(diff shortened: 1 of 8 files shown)


src/storage/store/account/message.rs
```diff
@@ -156,14 +156,35 @@ pub fn make_cast_id_key(cast_id: &CastId) -> Vec<u8> {
...
-    match db.get(&key)? {
+// We don't commit to the db until the end of the transaction, so, for cases where we might be handling conflicting messages within the same transaction,
+// We need to check against the transaction batch first. e.g. A cast add and a cast remove for the same cast_id in the same transaction should not both be merged
+pub fn get_from_db_or_txn(
+    db: &RocksDB,
+    txn: &mut RocksDbTransactionBatch,
+    key: &[u8],
+) -> Result<Option<Vec<u8>>, HubError> {
+    if let Some(value) = txn.batch.get(key) {
+        Ok(value.clone())
```
sanjayprabhu, line 176: I think this might not correctly handle deletes that were made to the transaction. We might still be falling back to the db.

sanjayprabhu: Added a test, and confirmed it's working as intended. The batch stores an Option, and None counts as a cache hit so the db is not consulted.

PR #171, opened 2024-12-17
turbohash

suurkivi:
This change stores the hash of a trie node in the node's parent (in a hashmap).

This yields an important performance improvement when recomputing the hash of a changed node. Previously, when recomputing the hash, as we are working our way up the trie, we had to load siblings of a changed node out of the db. In this version, the sibling hashes are already loaded, which dramatically reduces the number of db reads.

### DIFF
--- a/src/storage/trie/trie_node.rs
+++ b/src/storage/trie/trie_node.rs
@@ -58,21 +58,20 @@ impl<'a> Drop for Context<'a> {
 /// and keeps track of the number of items in the subtree.
 #[derive(Default, Debug, Clone)]
 pub struct TrieNode {
-    hash: Vec<u8>,
     items: usize,
     children: HashMap<u8, TrieNodeType>,
+    child_hashes: HashMap<u8, Vec<u8>>,
     key: Option<Vec<u8>>,
 }
 
 // An empty struct that represents a serialized trie node, which will need to be loaded from the db
+// TODO: since SerializedTrieNode is now empty, do we need it at all?
 #[derive(Debug, Clone)]
-pub struct SerializedTrieNode {
-    pub hash: Option<Vec<u8>>,
-}
+pub struct SerializedTrieNode {}
 
 impl SerializedTrieNode {
-    fn new(hash: Option<Vec<u8>>) -> Self {
-        SerializedTrieNode { hash }
+    fn new() -> Self {
+        SerializedTrieNode {}
     }
 }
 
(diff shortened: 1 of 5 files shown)


src/storage/trie/trie_node_tests.rs
```diff
@@ -217,14 +218,14 @@ mod tests {
...
+        let r = node.delete(ctx, &mut hm(), &db, &mut txn, vec![key1.clone()], 0);
         assert_eq!(r.unwrap()[0], true);
         assert_eq!(node.items(), 1);
 
-        let r = node.delete(ctx, &db, &mut txn, vec![key2.clone()], 0);
+        let r = node.delete(ctx, &mut hm(), &db, &mut txn, vec![key2.clone()], 0);
         assert_eq!(r.unwrap()[0], true);
         assert_eq!(node.items(), 0);
-        assert_eq!(node.hash(), empty_hash());
+        assert_eq!(node.hash(), Vec::<u8>::new());
```
suurkivi, line 228: note a subtle change here

src/storage/trie/trie_node_tests.rs
```diff
@@ -251,14 +252,14 @@ mod tests {
...
+        let r = node.insert(ctx, &mut hm(), &db, &mut txn, vec![key.clone()], 0);
         assert_eq!(r.unwrap()[0], true);
 
         // delete the key
-        let r = node.delete(ctx, &db, &mut txn, vec![key.clone()], 0);
+        let r = node.delete(ctx, &mut hm(), &db, &mut txn, vec![key.clone()], 0);
         assert_eq!(r.unwrap()[0], true);
         assert_eq!(node.items(), 0);
-        assert_eq!(node.hash(), empty_hash());
+        assert_eq!(node.hash(), Vec::<u8>::new());
```
suurkivi, line 262: note a subtle change here

src/storage/trie/trie_node.rs
```diff
@@ -122,15 +124,25 @@ impl TrieNode {
         for char in db_trie_node.child_chars {
             children.insert(
                 char as u8,
-                TrieNodeType::Serialized(SerializedTrieNode::new(None)),
+                TrieNodeType::Serialized(SerializedTrieNode::new()),
```
sanjayprabhu, line 127: If the struct is empty, do we actually need the struct?

suurkivi: left a TODO comment

PR #173, opened 2024-12-18
support fname signature verifications

CassOnMars:


src/storage/store/engine.rs
```diff
@@ -444,11 +444,17 @@ impl ShardEngine {
                         "Fname transfer has no proof"
                     );
                 }
-                let proof = fname_transfer.proof.as_ref().unwrap();
-                // TODO: Verify the EIP-712 server signature
+
+                match fname_transfer.verify_signature() {
```
sanjayprabhu, line 448: We should also do the same thing here https://github.com/farcasterxyz/snapchain-v0/blob/f5f1aff56fc11f9fb8421e6eeab9277628e35998/src/storage/store/engine.rs#L853-L855 for username proof claim signatures (I just looked and looks like we never call `verifyUserNameProofClaim` in hubs, this might be a bug?) and the eth verified addresses.

Could you also implement sol verified addresses (maybe in a separate PR? upto you) before moving on to the other validations?

src/core/message.rs
```diff
@@ -40,3 +45,151 @@ impl proto::ValidatorMessage {
...
+    pub fn verify_signature(&self) -> Result<bool, HubError> {
+        let proof = self.proof.as_ref().unwrap();
+        let username = std::str::from_utf8(&proof.name);
+        if username.is_err() {
+            return Err(HubError::validation_failure(
+                "could not deserialize username",
+            ));
+        }
+
+        let json = json!({
```
sanjayprabhu, line 59: We should make this a const inside validations.ts

sanjayprabhu (commented):
Nice

CassOnMars:
The existing test failing is correct for failing, this test has invalid parameters, but we should add helpers to streamline generating valid signed payloads for tests instead.

PR #187, opened 2024-12-24
Switch to Malachite crate

sds:
Now that Malachite has been released as a crate, we no longer need to install it via Git.

### DIFF
--- a/src/consensus/consensus.rs
+++ b/src/consensus/consensus.rs
@@ -11,7 +11,7 @@ use tracing::{debug, error, info, warn};
 
 use informalsystems_malachitebft_config::TimeoutConfig;
 use informalsystems_malachitebft_core_consensus::{
-    Effect, ProposedValue, Resume, SignedConsensusMsg, ValueToPropose,
+    Effect, ProposedValue, Resumable, Resume, SignedConsensusMsg, ValueToPropose,
 };
 use informalsystems_malachitebft_core_types::{
     Context, Extension, Round, SignedProposal, SignedProposalPart, SignedVote, Timeout, TimeoutKind,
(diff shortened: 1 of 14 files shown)


adizere:
Remember to update the main readme.md (and any other docs) before merging this.

src/consensus/validator.rs
```diff
@@ -134,7 +134,8 @@ impl ShardValidator {
         ProposedValue {
             height: full_proposal.height(),
             round: full_proposal.round(),
-            validator_address: full_proposal.proposer_address(),
+            valid_round: full_proposal.round(),
```
sanjayprabhu, line 137: What's the different between round and valid_round?

src/consensus/consensus.rs
```diff
@@ -517,34 +621,8 @@ impl Consensus {
...
-                            .await?;
-                    }
-                }
-
-                Ok(Resume::Continue)
-            }
-
-            Effect::GetValue(height, round, timeout) => {
-                let timeout = timeouts.duration_for(timeout.step);
+            Effect::GetValue(height, round, timeout, _) => {
```
sanjayprabhu, line 624: There are a lot of these ignored params, are they all safe to ignore?

aditiharini: Good point. I rewrote the code to use the extra param-- the code still does the same thing, the extra param is a more typesafe way of handling these resume values. It makes sure that the correct enum is used for the resume without us having to specify.

src/consensus/validator.rs
aditiharini: valid_round is the POL round. I think it should be set to `Nil` for all the cases in our code. It should be populated only for cases where we're restreaming an old proposal, not when we're proposing for the first time.

sanjayprabhu (approved):

PR #197, opened 2025-01-06
add logging to make debugging trie mismatches easier

aditiharini:
Add logging for debugging trie root mismatches.

### DIFF
--- a/src/storage/store/engine.rs
+++ b/src/storage/store/engine.rs
@@ -349,6 +349,52 @@ impl ShardEngine {
         result
     }
 
+    fn txn_summary(txn: &Transaction) -> String {
+        let mut summary = String::new();
+        for message in &txn.user_messages {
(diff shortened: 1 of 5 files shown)


src/storage/store/engine.rs
```diff
@@ -349,6 +349,52 @@ impl ShardEngine {
         result
     }
 
+    fn txn_summary(txn: &Transaction) -> String {
+        let mut summary = String::new();
```
sanjayprabhu, line 353: I think the summary should include the fid once (each txn is per fid), and the identifier for each message type (message hash, txn hash and log index for onchain event and fname), every thing else we can determine based on the identifier.

aditiharini: Fixed. I included some of the redundant info because I think it'll be useful for eyeballing the logs and seeing if there are patterns.

Issue #180, opened 2024-12-21
Implement a mempool


sanjayprabhu:
Right now messages are only accepted by a single node andw e have to wait for that node's turn to be a proposer to have them be accepted. So, fix this and to allow eventual read nodes to submit messages, we should implement a gossip based mempool. It should:

- On receiving a message by submitMessage, the node should add it to a local hashmap and then gossip it out in a separate topic
- Other nodes should listen to the topic and add it to their local hashmap
- When messages are successfully included in the block, they should be evicted from the local hashmaps on all nodes
- The mempool should also enforce rate limits by fid based on the last n blocks (say, 5-10% of fid storage limit per 1000 blocks)
- The mempool should be persisted to disk to messages not lost on restart or crash

FrederikBolding:
@sanjayprabhu I have been looking at this a bit. How should the proposed hashmap mempool work with the existing ["mempool"](https://github.com/farcasterxyz/snapchain/blob/main/src/storage/store/engine.rs#L216) (meaning the message queue)? Should this be effectively a replacement or something else?

sanjayprabhu:
Should effectively be a replacement. 

I think we should create a separate topic for separation of concerns. Consensus messages are critical and we shouldn't have any other messages sharing that topic. Eventually I think we'd also want to enforce nodes that aren't active validators cannot publish to the consensus topic (if they do, it's treated as an invalid message and is not forwarded to the mesh).

Read nodes, when they are implemented, will want to listen and publish to the mempool topic but don't need to know about the consensus topic.

FrederikBolding:
Sounds good! I have started working on this, will let you know once I have something to share for feedback!

sanjayprabhu:
This is partially done. I'll file separate tickets for rate limits and persistence

PR #200, opened 2025-01-08
feat: Add gossip and eviction to the mempool

FrederikBolding:
This PR adds a basic implementation of the gossipped mempool described in #180.

It uses the existing mempool implementation but changes the key to use a key derived from the timestamp and identity of a message. Additions to the mempool are broadcast using a new gossip message. Messages are also evicted from the mempool when shard chunks are emitted that contain messages contained in the mempool.

Some minor changes to the channels and overall plumbing was required to setup the mempool communication.

### DIFF
--- a/src/core/message.rs
+++ b/src/core/message.rs
@@ -1,4 +1,3 @@
-use crate::mempool::mempool::MempoolKey;
 use crate::proto;
 use crate::proto::MessageType;
 
@@ -26,13 +25,6 @@ impl proto::Message {
     pub fn hex_hash(&self) -> String {
         hex::encode(&self.hash)
     }
-
-    pub fn mempool_key(&self) -> MempoolKey {
-        if let Some(data) = &self.data {
-            return MempoolKey::new(data.timestamp as u64, self.hex_hash());
-        }
-        todo!();
-    }
 }
 
 impl proto::ValidatorMessage {
@@ -46,20 +38,5 @@ impl proto::ValidatorMessage {
             return event.fid;
         }
         0
-    }
-
-    pub fn mempool_key(&self) -> MempoolKey {
-        if let Some(fname) = &self.fname_transfer {
-            if let Some(proof) = &fname.proof {
-                return MempoolKey::new(proof.timestamp, fname.id.to_string());
-            }
-        }
-        if let Some(event) = &self.on_chain_event {
-            return MempoolKey::new(
-                event.block_timestamp,
-                hex::encode(&event.transaction_hash) + &event.log_index.to_string(),
-            );
-        }
-        todo!();
     }
 }
--- a/src/core/message.rs
+++ b/src/core/message.rs
@@ -1,3 +1,4 @@
+use crate::mempool::mempool::MempoolKey;
 use crate::proto;
 use crate::proto::MessageType;
 
@@ -25,6 +26,13 @@ impl proto::Message {
     pub fn hex_hash(&self) -> String {
         hex::encode(&self.hash)
     }
+
+    pub fn mempool_key(&self) -> MempoolKey {
+        if let Some(data) = &self.data {
+            return MempoolKey::new(data.timestamp as u64, self.hex_hash());
+        }
+        todo!();
+    }
 }
 
 impl proto::ValidatorMessage {
@@ -40,16 +48,18 @@ impl proto::ValidatorMessage {
         0
     }
 
-    // TODO: Determine if these hashes are good enough for the mempool.
-    pub fn hex_hash(&self) -> String {
+    pub fn mempool_key(&self) -> MempoolKey {
         if let Some(fname) = &self.fname_transfer {
             if let Some(proof) = &fname.proof {
-                return hex::encode(&proof.signature);
+                return MempoolKey::new(proof.timestamp, fname.id.to_string());
             }
         }
         if let Some(event) = &self.on_chain_event {
-            return hex::encode(&event.transaction_hash);
+            return MempoolKey::new(
+                event.block_timestamp,
+                hex::encode(&event.transaction_hash) + &event.log_index.to_string(),
+            );
         }
-        todo!("Not implemented");
+        todo!();
     }
 }
--- a/src/core/message.rs
+++ b/src/core/message.rs
@@ -39,4 +39,17 @@ impl proto::ValidatorMessage {
         }
         0
     }
+
+    // TODO: Determine if these hashes are good enough for the mempool.
+    pub fn hex_hash(&self) -> String {
+        if let Some(fname) = &self.fname_transfer {
+            if let Some(proof) = &fname.proof {
+                return hex::encode(&proof.signature);
+            }
+        }
+        if let Some(event) = &self.on_chain_event {
+            return hex::encode(&event.transaction_hash);
+        }
+        todo!("Not implemented");
+    }
 }
--- a/src/main.rs
+++ b/src/main.rs
@@ -160,6 +160,7 @@ async fn main() -> Result<(), Box<dyn Error>> {
         node.shard_stores.clone(),
         gossip_tx.clone(),
         shard_decision_rx,
+        statsd_client.clone(),
     );
     tokio::spawn(async move { mempool.run().await });
 
--- a/src/main.rs
+++ b/src/main.rs
@@ -21,7 +21,7 @@ use std::net::SocketAddr;
 use std::process;
 use std::time::Duration;
 use tokio::signal::ctrl_c;
-use tokio::sync::mpsc;
+use tokio::sync::{broadcast, mpsc};
 use tokio::{select, time};
 use tonic::transport::Server;
 use tracing::{error, info, warn};
@@ -135,11 +135,14 @@ async fn main() -> Result<(), Box<dyn Error>> {
     let _ = Metrics::register(registry);
 
     let (messages_request_tx, messages_request_rx) = mpsc::channel(100);
+    let (shard_decision_tx, shard_decision_rx) = broadcast::channel(100);
+
     let node = SnapchainNode::create(
         keypair.clone(),
         app_config.consensus.clone(),
         Some(app_config.rpc_address.clone()),
         gossip_tx.clone(),
+        shard_decision_tx,
         None,
         messages_request_tx,
         block_store.clone(),
@@ -155,7 +158,8 @@ async fn main() -> Result<(), Box<dyn Error>> {
         messages_request_rx,
         app_config.consensus.num_shards,
         node.shard_stores.clone(),
-        Some(gossip_tx.clone()),
+        gossip_tx.clone(),
+        shard_decision_rx,
     );
     tokio::spawn(async move { mempool.run().await });
 
--- a/src/main.rs
+++ b/src/main.rs
@@ -105,9 +105,14 @@ async fn main() -> Result<(), Box<dyn Error>> {
     );
 
     let (system_tx, mut system_rx) = mpsc::channel::<SystemMessage>(100);
+    let (mempool_tx, mempool_rx) = mpsc::channel(app_config.mempool.queue_size as usize);
 
-    let gossip_result =
-        SnapchainGossip::create(keypair.clone(), app_config.gossip, system_tx.clone());
+    let gossip_result = SnapchainGossip::create(
+        keypair.clone(),
+        app_config.gossip,
+        system_tx.clone(),
+        mempool_tx.clone(),
+    );
 
     if let Err(e) = gossip_result {
         error!(error = ?e, "Failed to create SnapchainGossip");
@@ -144,8 +149,8 @@ async fn main() -> Result<(), Box<dyn Error>> {
     )
     .await;
 
-    let (mempool_tx, mempool_rx) = mpsc::channel(app_config.mempool.queue_size as usize);
     let mut mempool = Mempool::new(
+        1024,
         mempool_rx,
         messages_request_rx,
         app_config.consensus.num_shards,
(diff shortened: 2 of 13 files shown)


src/main.rs
```diff
@@ -204,6 +214,43 @@ async fn main() -> Result<(), Box<dyn Error>> {
...
+            if let Some(mempool_message) = mempool_rx.recv().await {
+                match mempool_message {
+                    MempoolMessage::UserMessage(message) => {
+                        let result = service.submit_message_internal(message, false).await;
+                        if let Err(e) = result {
+                            warn!("Failed to add message to mempool: {:?}", e);
+                        }
+                    }
+                    MempoolMessage::ValidatorMessage(_validator_message) => {
+                        // TODO
```
FrederikBolding, line 247: I'm not sure whether these should even be gossipped and included in the mempool or not, since they come from the "connectors"?

src/main.rs
```diff
@@ -204,6 +214,43 @@ async fn main() -> Result<(), Box<dyn Error>> {
...
+            app_config.consensus.num_shards,
+            Box::new(routing::ShardRouter {}),
+            l1_client,
+        );
+
+        loop {
+            if let Some(mempool_message) = mempool_rx.recv().await {
+                match mempool_message {
+                    MempoolMessage::UserMessage(message) => {
+                        let result = service.submit_message_internal(message, false).await;
```
FrederikBolding, line 241: This seemed like the most sane way to share the validation logic between RPC-submitted messages and gossip-submitted messages

sanjayprabhu: The dependency is inverted. The mempool should live one level below the service, so it shouldn't have any knowledge of the service. The service should be submitting to the mempool. See the new structure in https://github.com/farcasterxyz/snapchain/pull/203

src/main.rs
sanjayprabhu: No, these should not be gossiped

src/mempool/mempool.rs
```diff
@@ -1,4 +1,70 @@
...
+            gossip_tx,
+        }
+    }
+
+    pub async fn insert(
+        &mut self,
+        message: MempoolMessage,
+    ) -> Result<(), SendError<GossipEvent<SnapchainValidatorContext>>> {
+        if self.mempool.len() >= self.capacity {
+            // TODO: Throw
```
sanjayprabhu, line 35: Should probably just log and ignore?

src/storage/store/engine.rs
```diff
@@ -212,11 +241,16 @@ impl ShardEngine {
...
 
-            while messages.len() < self.max_messages_per_block as usize {
-                match self.messages_rx.try_recv() {
-                    Ok(msg) => messages.push(msg),
-                    Err(mpsc::error::TryRecvError::Empty) => break,
-                    Err(err) => return Err(EngineError::from(err)),
+            let capacity = self.max_messages_per_block as usize - messages.len();
+
+            if capacity > 0 {
+                let mempool = self.mempool.read().await;
```
sanjayprabhu, line 247: Our philosophy has been to avoid locks and prefer channels for cross thread communication.

src/main.rs
```diff
@@ -294,6 +341,9 @@ async fn main() -> Result<(), Box<dyn Error>> {
                     }
                 }
             }
+            Some(block) = block_rx.recv() => {
+                node.handle_block(block).await;
```
sanjayprabhu, line 345: We should do this at the shard level, and not the block level because shards will be making independent progress. And waiting for block confirmation before eviction would be too slow.

src/node/snapchain_node.rs
```diff
@@ -223,4 +229,24 @@ impl SnapchainNode {
...
+            let header = chunk.header.expect("Expects chunk to have a header");
+            let height = header.height.expect("Expects header to have a height");
+            let mempool = self
+                .shard_mempools
+                .get(&height.shard_index)
+                .expect("Expects mempool to exist for shard");
+            let mut mempool_write = mempool.write().await;
+            for transaction in chunk.transactions {
+                for user_message in transaction.user_messages {
+                    mempool_write.remove(user_message.hex_hash());
```
sanjayprabhu, line 244: See note about locks. I would structure this as one single channel for receiving confirmed shard chunks that's then published to a channel that the mempool listens too. In a select loop within the mempool, evict the messages that were included in the block. 

The other thing to consider is how to evict messages that we rejected before being included in the block (e.g. validation failed or rate limits were exceeded etc). We the mempool channel might need to be more generic than just accepting shard chunks.

src/proto/blocks.proto
```diff
@@ -140,11 +140,19 @@ message RegisterValidator {
   uint64 nonce = 2;
 }
 
+message MempoolMessage {
+  oneof mempool_message {
+    ValidatorMessage validator_message = 1;
```
sanjayprabhu, line 145: Validator messages should not be gossiped.

src/mempool/mempool.rs
```diff
@@ -25,40 +27,44 @@ impl Default for Config {
...
 
 pub struct Mempool {
+    capacity_per_shard: usize,
     shard_stores: HashMap<u32, Stores>,
     message_router: Box<dyn MessageRouter>,
     num_shards: u32,
     mempool_rx: mpsc::Receiver<MempoolMessage>,
     messages_request_rx: mpsc::Receiver<MempoolMessagesRequest>,
-    messages: HashMap<u32, BTreeMap<MempoolKey, MempoolMessage>>,
+    messages: HashMap<u32, BTreeMap<String, MempoolMessage>>,
```
FrederikBolding, line 43: This changes the ordering in which messages are pulled from the mempool, if the ordering is important (e.g. if we want FIFO) we may need to find another way to choose messages from the pool

src/mempool/mempool.rs
```diff
@@ -149,6 +170,43 @@ impl Mempool {
         return true;
     }
 
+    async fn insert(&mut self, message: MempoolMessage) {
```
sanjayprabhu, line 173: Let's emit some stats here, number of inserts tagged by success/failure and message type, and a guage for the current size of the mempool by shard.

src/mempool/mempool.rs
```diff
@@ -159,26 +217,27 @@ impl Mempool {
...
+                        let height = header.height.expect("Expects header to have a height");
+                        if let Some(mempool) = self.messages.get_mut(&height.shard_index) {
+                            for transaction in chunk.transactions {
+                                for user_message in transaction.user_messages {
+                                    mempool.remove(&user_message.mempool_key());
                                 }
-                                Some(messages) => {
-                                    messages.insert(MempoolKey { inserted_at: Instant::now()}, message.clone());
+                                for system_message in transaction.system_messages {
+                                    mempool.remove(&system_message.mempool_key());
```
sanjayprabhu, line 230: We should record the same stats as above when removing from the mempool as well

sanjayprabhu (commented):
Nice, generally looks good. Please add some tests for:
 - Messages being evicted from the mempool on confirmation
 - Messages submitted to one node is gossiped to another node.

src/mempool/mempool.rs
```diff
@@ -160,6 +220,55 @@ impl Mempool {
...
+            match self.messages.get_mut(&shard_id) {
+                None => {
+                    let mut messages = BTreeMap::new();
+                    messages.insert(message.mempool_key(), message.clone());
+                    self.messages.insert(shard_id, messages);
+                    self.statsd_client
+                        .gauge_with_shard(shard_id, "mempool.size", 1);
+                }
+                Some(messages) => {
+                    if messages.len() >= self.capacity_per_shard {
```
aditiharini, line 238: I think it'd be good to add some logging here so it's easier to debug cases where messages that were successfully submitted didn't make it onto the network. Having fid in a log tag would be useful.

src/mempool/mempool.rs
```diff
@@ -34,7 +37,52 @@ impl Default for Config {
...
+            timestamp,
+            identity,
+        }
+    }
+}
+
+impl proto::Message {
+    pub fn mempool_key(&self) -> MempoolKey {
+        if let Some(data) = &self.data {
+            return MempoolKey::new(data.timestamp as u64, self.hex_hash());
```
aditiharini, line 56: I think we should use the arrival timestamp and not the timestamp on the message itself for ordering. If we use the timestamp on the message, backdated messages will always be prioritized.

src/proto/blocks.proto
```diff
@@ -140,11 +140,18 @@ message RegisterValidator {
   uint64 nonce = 2;
 }
 
+message MempoolMessage {
+  oneof mempool_message {
```
aditiharini, line 144: Why is this a oneof?

src/mempool/mempool.rs
```diff
@@ -172,19 +281,22 @@ impl Mempool {
...
-                            // TODO(aditi): We need a size limit on the mempool and we need to figure out what to do if it's exceeded
-                            match self.messages.get_mut(&shard_id) {
-                                None => {
-                                    let mut messages = BTreeMap::new();
-                                    messages.insert(MempoolKey { inserted_at: Instant::now()}, message.clone());
-                                    self.messages.insert(shard_id, messages);
+                        self.insert(message).await;
+                    }
+                }
+                chunk = self.shard_decision_rx.recv() => {
```
aditiharini, line 287: Is there a reason to use the shard decision rx over the block decision rx?

src/proto/blocks.proto
FrederikBolding: Ah, this is leftover from my original commit when I had validation messages gossipped as well.

I can remove, but I was wondering if it is simpler to expand it in the future like this?

src/mempool/mempool.rs
FrederikBolding: I did this at the block level at first, but @sanjayprabhu recommended doing it at the shard level instead here: https://github.com/farcasterxyz/snapchain/pull/200#discussion_r1908215651

src/mempool/mempool.rs
FrederikBolding:
```suggestion
                    if messages.len() >= self.capacity_per_shard {
                    warn!(fid = message.fid(), "Message dropped due to mempool being over capacity");
```
Something along these lines?

src/mempool/mempool.rs
FrederikBolding: https://github.com/farcasterxyz/snapchain/pull/200/commits/4842eafcd741eb9c951b24b86085989ad9280544

src/mempool/mempool.rs
FrederikBolding: https://github.com/farcasterxyz/snapchain/pull/200/commits/4842eafcd741eb9c951b24b86085989ad9280544

FrederikBolding: Let me know if you want additional stats or tweaks to the formatting

src/core/message.rs
```diff
@@ -25,6 +26,13 @@ impl proto::Message {
     pub fn hex_hash(&self) -> String {
         hex::encode(&self.hash)
     }
+
+    pub fn mempool_key(&self) -> MempoolKey {
+        if let Some(data) = &self.data {
+            return MempoolKey::new(data.timestamp as u64, self.hex_hash());
```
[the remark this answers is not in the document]
FrederikBolding: Moved things around a bit in https://github.com/farcasterxyz/snapchain/pull/200/commits/4842eafcd741eb9c951b24b86085989ad9280544 - lmk what you think!

src/mempool/mempool.rs
FrederikBolding: The existing implementation is FIFO and so I was trying to preserve that behaviour, will leave a comment.

src/mempool/mempool.rs
FrederikBolding: https://github.com/farcasterxyz/snapchain/pull/200/commits/118c035be8bc035f2d262a5cbb23553651a3e8dc

src/mempool/mempool.rs
FrederikBolding: https://github.com/farcasterxyz/snapchain/pull/200/commits/118c035be8bc035f2d262a5cbb23553651a3e8dc

src/mempool/mempool.rs
aditiharini: Yes, can you add the hash of the message or some identifier too?

src/mempool/mempool.rs
aditiharini: Got it, that reasoning makes sense. There are minimal chances that a message is committed at the shard level but not the block level.

src/mempool/mempool_test.rs
```diff
@@ -97,4 +122,113 @@ mod tests {
         }));
         assert!(!valid)
     }
+
+    const HOST_FOR_TEST: &str = "127.0.0.1";
+    const PORT_FOR_TEST: u32 = 9388;
+
+    fn setup_gossip_mempool(
```
aditiharini, line 129: Why are there 2 different setup functions? Do the existing tests break with the gossip set up?

src/mempool/mempool.rs
FrederikBolding: Done!

src/mempool/mempool_test.rs
aditiharini: I think it makes sense to set up the full mempool with gossip for all tests.

aditiharini: If you find that the tests are getting out of hand with this approach, let's reconsider though.

src/proto/blocks.proto
aditiharini: Hm leaving as oneof for extensibility seems reasonable

src/mempool/mempool_test.rs
FrederikBolding: https://github.com/farcasterxyz/snapchain/pull/200/commits/da9e455b4a4df918e775aec1aaca28ead0881d8b

sanjayprabhu (approved):
Nice, looks good to me.

PR #203, opened 2025-01-08
pull messages out of mempool rather than pushing messages into engine

aditiharini:
Pull messages out of the mempool and run validations at that point so all messages fed to the engine are likely to be valid.

### DIFF
--- a/src/mempool/mempool.rs
+++ b/src/mempool/mempool.rs
@@ -1,7 +1,10 @@
 use std::collections::{BTreeMap, HashMap};
 
 use serde::{Deserialize, Serialize};
-use tokio::sync::{mpsc, oneshot};
+use tokio::{
+    sync::{mpsc, oneshot},
+    time::Instant,
+};
 
 use crate::storage::{
     store::{engine::MempoolMessage, stores::Stores},
@@ -24,7 +27,7 @@ impl Default for Config {
 
 #[derive(Debug, PartialEq, Eq, PartialOrd, Ord)]
 pub struct MempoolKey {
-    trie_key: Vec<u8>,
+    inserted_at: Instant,
 }
 
 pub struct Mempool {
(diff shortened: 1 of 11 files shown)


src/storage/store/engine.rs
```diff
@@ -203,30 +196,42 @@ impl ShardEngine {
...
+                            err.to_string()
+                        )
+                    }
 
-            while messages.len() < self.max_messages_per_block as usize {
-                match self.messages_rx.try_recv() {
-                    Ok(msg) => messages.push(msg),
-                    Err(mpsc::error::TryRecvError::Empty) => break,
-                    Err(err) => return Err(EngineError::from(err)),
+                    match message_rx.await {
```
sanjayprabhu, line 218: I think we should sending in the max_messages_per_block and get all the messages in one go. You can also combine it with the timeout function to get rid of the loop https://docs.rs/tokio/latest/tokio/time/fn.timeout.html

src/mempool/mempool.rs
```diff
@@ -110,21 +145,33 @@ impl Mempool {
...
-                match senders {
-                    None => {
-                        error!("Unable to find shard to send message to")
-                    }
-                    Some(senders) => {
-                        if let Err(err) = senders.messages_tx.send(message).await {
-                            error!("Unable to send message to engine: {}", err.to_string())
+        loop {
+            tokio::select! {
+                message = self.mempool_rx.recv() => {
```
sanjayprabhu, line 150: Selecting directly here is dangerous, the mempool rx channel and overload the messages_request_rx channel, and make it so the request is never fulfilled if we get sufficient message traffic to the mempool.

It's better to loop using a tick and ensure that the messages request channel is processed first and always prioritized. 

See https://github.com/farcasterxyz/snapchain/blob/41f1e05950a216d57fac04c56d27f3830fea0fa5/src/consensus/proposer.rs#L284-L288 for an example of how we're done this elsewhere.

aditiharini: I made a change to specify `biased` in the select macro. This causes the macro to preference branches in order of specification. I think this is what we want here? 

In this case, I don't really think we need a timeout because we will just poll indefinitely.

src/storage/store/engine.rs
aditiharini: Agreed-- changed it so we return a bulk set of messages in a single response. 

I just left the loop in for now. I think in order for the timeout to work, we need to pass a timeout into the mempool via the messages request and have the mempool timeout the request after the timeout has been hit. Is that what you were imagining? I don't think we'll get the behavior we want with a timeout on the engine side.

src/mempool/mempool.rs
```diff
@@ -75,32 +82,60 @@ impl Mempool {
...
+                                break;
+                            }
+                        }
+                    };
+                }
             }
+        }
+
+        if let Err(_) = tx.send(message) {
+            error!("Unable to send message from mempool");
```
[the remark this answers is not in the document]
aditiharini: The error just has the entire message inside of it.

PR #215, opened 2025-01-14
fix handling of ens usernames

aditiharini:
Look up ens proofs in the username proofs store rather than the user data store.

### DIFF
--- a/src/storage/store/engine.rs
+++ b/src/storage/store/engine.rs
@@ -1,13 +1,13 @@
+use super::account::UsernameProofStore;
 use super::account::{IntoU8, OnchainEventStorageError, UserDataStore};
 use crate::core::error::HubError;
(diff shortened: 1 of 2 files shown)


src/storage/store/engine.rs
```diff
@@ -963,32 +964,67 @@ impl ShardEngine {
...
+                            }
+                            _ => Ok(None),
+                        },
+                        None => Ok(None),
+                    },
+                },
+                None => Ok(None),
+            }
+        } else {
+            UserDataStore::get_username_proof(&self.stores.user_data_store, txn, name.as_bytes())
```
sanjayprabhu, line 999: Let us add a check in the userdata store to panic if the name ends with .eth, and a todo to move it to the username proof store.

sanjayprabhu (approved):

PR #218, opened 2025-01-15
use the message store for mempool validation

aditiharini:
Switch from using the trie to the account store for the duplicate message check in the mempool.

### DIFF
--- a/src/mempool/mempool.rs
+++ b/src/mempool/mempool.rs
@@ -73,6 +73,7 @@ impl Mempool {
         let fid = message.fid();
         let shard = self.message_router.route_message(fid, self.num_shards);
         let stores = self.shard_stores.get_mut(&shard);
+        // Default to false in the orror paths
         match stores {
             None => {
                 error!("Error finding store for shard: {}", shard);
--- a/src/mempool/mempool.rs
+++ b/src/mempool/mempool.rs
@@ -62,7 +62,8 @@ impl Mempool {
         }
     }
 
-    fn message_exists_in_trie(&mut self, fid: u64, trie_key: Vec<u8>) -> bool {
+    fn message_already_exists(&mut self, message: &MempoolMessage) -> bool {
+        let fid = message.fid();
         let shard = self.message_router.route_message(fid, self.num_shards);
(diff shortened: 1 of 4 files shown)


src/mempool/mempool.rs
```diff
@@ -62,29 +69,62 @@ impl Mempool {
...
-                        false
+            Some(stores) => match message {
+                MempoolMessage::UserMessage(message) => match &message.data {
+                    None => false,
+                    Some(message_data) => {
+                        let ts_hash = make_ts_hash(message_data.timestamp, &message.hash).unwrap();
+                        let set_postfix = type_to_set_postfix(message_data.r#type());
+                        let primary_key =
+                            make_message_primary_key(fid, set_postfix as u8, Some(&ts_hash));
+                        let existing_message = get_message_by_key(
```
sanjayprabhu, line 90: This is too low level for the mempool. It should just be doing `stores.message_exists(&message);`

src/mempool/mempool.rs
```diff
@@ -62,29 +69,62 @@ impl Mempool {
...
+                        );
+                        match existing_message {
+                            Ok(Some(_)) => true,
+                            Err(_) | Ok(None) => false,
+                        }
+                    }
+                },
+                MempoolMessage::ValidatorMessage(message) => {
+                    if let Some(onchain_event) = &message.on_chain_event {
+                        match stores.onchain_event_store.exists(&onchain_event) {
```
sanjayprabhu, line 103: I would push this into `stores.onchain_event_exists` as well.

src/mempool/mempool.rs
```diff
@@ -62,29 +69,62 @@ impl Mempool {
...
+                            Err(_) => return false,
+                            Ok(exists) => return exists,
+                        }
+                    }
+
+                    if let Some(fname_transfer) = &message.fname_transfer {
+                        match &fname_transfer.proof {
+                            None => return false,
+                            Some(proof) => {
+                                let username_proof = UserDataStore::get_username_proof(
```
sanjayprabhu, line 113: Same here

PR #219, opened 2025-01-16
fix: incorrect use of encode_to_vec() method

offensif:
The String type does not have an implementation of the encode_to_vec() method. Better use the as_bytes().to_vec() method to convert a string into a vector of bytes


offensif:
@aditiharini Hello, my Pull Request is ready for review!

PR #226, opened 2025-01-23
feat: support validations for different types, refactor

CassOnMars:


src/core/validations/cast_test.rs
```diff
@@ -0,0 +1,98 @@
...
+    #[derive(Deserialize)]
+    struct PagedResponse {
+        messages: Vec<Message>,
+    }
+
+    #[tokio::test]
+    async fn test_cast_validation() {
+        let n: u32 = rand::thread_rng().gen::<u32>() % 10000;
+        let resp = reqwest::get(format!(
+            "https://hoyt.farcaster.xyz:2281/v1/castsByFid?fid={}",
```
sanjayprabhu, line 51: This might end up being flaky. If it is, I'd add this as a separate test that's run manually and CI uses a known set of messages checked into the repo.

src/core/validations/mod.rs
```diff
@@ -0,0 +1,60 @@
...
+    if url_bytes.is_empty() {
+        return Err(ValidationError::InvalidDataLength);
+    }
+
+    if url_bytes.len() > 256 {
+        return Err(ValidationError::InvalidDataLength);
+    }
+
+    if !url.is_ascii() {
+        return Err(ValidationError::InvalidData);
```
sanjayprabhu, line 56: We might want to add a field name to these error codes otherwise it's going to be hard to figure out what exactly failed validation. I'll ticket this.

src/storage/store/engine.rs
```diff
@@ -947,15 +950,55 @@ impl ShardEngine {
                 // Validate ens
```
sanjayprabhu, line 950: What about this one?

src/storage/store/engine.rs
```diff
@@ -947,15 +950,55 @@ impl ShardEngine {
...
+            }
+            Some(proto::message_data::Body::LinkBody(link_body)) => {
+                let result = link::validate_link_body(link_body);
+                if result.is_err() {
+                    return Err(MessageValidationError::MessageValidationError(
+                        result.unwrap_err(),
+                    ));
+                }
             }
             _ => {}
```
sanjayprabhu, line 1003: We should remove this, so we know we've covered every case.

src/storage/store/engine.rs
```diff
@@ -947,15 +950,55 @@ impl ShardEngine {
                 // Validate ens
             }
             Some(proto::message_data::Body::VerificationAddAddressBody(add)) => {
-                let result = validations::validate_add_address(add, message_data.fid, network);
+                let result = verification::validate_add_address(add, message_data.fid, network);
```
sanjayprabhu, line 953: We should push all the branches of this match into `validations::message::validate_message` above. So that validate_message can be accessed independently via a validations crate, and not tied into the engine code. 

In hubble, these set of validations were for things that depended on external data or interdependencies between messages (e.g. validating ENS names had a match verification add)

sanjayprabhu (commented):
Generally looks good, the main thing I'd change is moving the validation of message body into `validate_message` and removing the catchall match clause. Other things can happen later.

src/storage/store/engine.rs
CassOnMars: this was pulled out because validation needs to happen at a higher level. did you want there to be just basic field info validation (minus the signature/address?)

src/storage/store/engine.rs
CassOnMars: There are validations dependent on existing state, e.g. `validate_username` that queries the db for additional info – do we _want_ this in the validations methods?

src/core/validations/message.rs
```diff
@@ -34,6 +48,67 @@ pub fn validate_message(message: &proto::Message) -> Result<(), ValidationError>
         &message.signer,
     )?;
 
+    let network = FarcasterNetwork::try_from(message_data.network)
+        .or_else(|_| Err(ValidationError::InvalidData))?;
+
+    match &message_data.body {
+        Some(proto::message_data::Body::UserDataBody(_)) => {
+            // Split user data validation from engine
```
sanjayprabhu, line 56: This should have all the length validations from https://github.com/farcasterxyz/hub-monorepo/blob/913c0f67665e31241e45400b6c4f7471f28f75e2/packages/core/src/validations.ts#L977

sanjayprabhu (commented):
Looks good to me, should be good to merge one the user data validations are in.

PR #230, opened 2025-01-26
include deleted messages in emitted hub event for link compactions

aditiharini:
The account stores and trie were inconsistent for link compaction messages because we don't emit `deleted_messages` and we update the trie based on the data in the event. This led to the storage usage calculations being inconsistent across hubs and snapchain.

### DIFF
--- a/src/network/server.rs
+++ b/src/network/server.rs
@@ -344,10 +344,7 @@ impl MyHubService {
                             }
                         }
                     }
-                    proto::hub_event::Body::PruneMessageBody(_)
-                    | proto::hub_event::Body::RevokeMessageBody(_)
-                    | proto::hub_event::Body::MergeUsernameProofBody(_)
-                    | proto::hub_event::Body::MergeOnChainEventBody(_) => {}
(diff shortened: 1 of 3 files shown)


src/network/server.rs
```diff
@@ -326,6 +327,33 @@ impl MyHubService {
...
+                                    // result in an unbounded message size:
+                                    merge_message_body.deleted_messages = vec![]
+                                }
+                            }
+                        }
+                    }
+                    proto::hub_event::Body::PruneMessageBody(_)
+                    | proto::hub_event::Body::RevokeMessageBody(_)
+                    | proto::hub_event::Body::MergeUsernameProofBody(_)
+                    | proto::hub_event::Body::MergeOnChainEventBody(_) => {}
```
sanjayprabhu, line 350: I think we can just use the fallback `_ => {}` here to cover all of these cases

sanjayprabhu (approved):

sanjayprabhu (approved):

PR #233, opened 2025-01-28
upload snapshots to s3

aditiharini:
Clear existing snapshots and upload new ones to s3 via an admin rpc.


src/network/admin_server.rs
```diff
@@ -76,12 +89,52 @@ pub enum AdminServiceError {
...
     }
+
+    async fn backup_shard(&self, shard_id: u32, db: Arc<RocksDB>, now: i64) -> Result<(), Status> {
+        // TODO(aditi): Eventually, we should upload a metadata file. For now, just clear all existing snapshots and only keep 1 snapshot per shard
+        clear_snapshots(self.fc_network, &self.snapshot_config, shard_id)
+            .await
+            .map_err(|err| Status::from_error(Box::new(err)))?;
+        let db_location = db.location();
+        let backup_dir = Path::new(&db_location)
+            .parent()
```
sanjayprabhu, line 117: What's the actual directory structure for the backups? I think we should create an explicit directory called `backup` and place the backup there.

src/network/admin_server.rs
```diff
@@ -76,12 +89,52 @@ pub enum AdminServiceError {
...
             db_manager,
             mempool_tx,
+            shard_stores,
+            block_store,
+            snapshot_config,
+            fc_network,
         }
     }
+
+    async fn backup_shard(&self, shard_id: u32, db: Arc<RocksDB>, now: i64) -> Result<(), Status> {
```
sanjayprabhu, line 110: We should call this backup_and_upload

src/network/admin_server.rs
```diff
@@ -76,12 +89,52 @@ pub enum AdminServiceError {
...
+        .map_err(|err| Status::from_error(Box::new(err)))?;
+        storage::db::snapshot::upload_to_s3(
+            self.fc_network,
+            tar_gz_path,
+            &self.snapshot_config,
+            shard_id,
+        )
+        .await
+        .map_err(|err| Status::from_error(Box::new(err)))?;
+        Ok(())
```
sanjayprabhu, line 136: We should delete the local files after successful upload (and maybe even on failure?)

src/main.rs
```diff
@@ -164,7 +164,14 @@ async fn main() -> Result<(), Box<dyn Error>> {
     );
     tokio::spawn(async move { mempool.run().await });
 
-    let admin_service = MyAdminService::new(db_manager, mempool_tx.clone());
+    let admin_service = MyAdminService::new(
```
sanjayprabhu, line 167: note, we're calling backup_db here in main. I think it's worth keeping that but maybe using an explicit config variable like backup_on_startup or similar.

src/network/admin_server.rs
```diff
@@ -178,4 +231,22 @@ impl AdminService for MyAdminService {
...
+            .duration_since(UNIX_EPOCH)
+            .map_err(|err| Status::from_error(Box::new(err)))?
+            .as_millis();
+
+        self.backup_shard(0, self.block_store.db.clone(), now as i64)
+            .await?;
+        for (shard, stores) in self.shard_stores.iter() {
+            self.backup_shard(*shard, stores.db.clone(), now as i64)
+                .await?
+        }
```
sanjayprabhu, line 249: Is this happening in the context of a request? Might want to return immediately and do the actual backup in a background thread.

Also, we should add protection for multiple requests. It shouldn't cause multiple backups (maybe just check for the presence of the backup dir?).

src/main.rs
aditiharini: What's the point of backing up on startup? Right now, there's very likely going to be nothing new in the db right?

aditiharini: I added the flag and left this code as-is. Left a comment that we might want to reconsider this, especially when we schedule snapshots.

src/network/admin_server.rs
aditiharini: I added protection for parallel requests. 

I'm not sure about returning early-- then we can't return the error back to the client. Maybe we want to turn this into a streaming rpc? I'll hold off on making this change-- if we really want it we can do it later.

src/network/admin_server.rs
aditiharini: Done, moved to `.rocks.backup`

PR #234, opened 2025-01-29
Malachite consensus rewrite

sanjayprabhu:
Upgrade malachite to 0.0.1 and re-write malachite consensus integration to work at the actor level so we can use WAL, Sync and the Consensus actors directly.


src/consensus/malachite/network_connector.rs
```diff
@@ -0,0 +1,332 @@
...
+
+pub struct MalachiteNetworkConnector<Codec> {
+    pub codec: Codec,
+}
+
+pub enum NetworkConnectorState {
+    Stopped,
+    Running {
+        // peers: BTreeSet<PeerId>,
+        output_port: OutputPort<NetworkEvent<SnapchainValidatorContext>>,
```
aditiharini, line 25: What's the point of output_port? What are we outputting to?

src/consensus/malachite/network_connector.rs
```diff
@@ -0,0 +1,332 @@
...
+        else {
+            return Ok(());
+        };
+
+        match msg {
+            Msg::Subscribe(subscriber) => {
+                subscriber.subscribe_to_port(output_port);
+            }
+
+            Msg::Publish(msg) => match msg {
```
aditiharini, line 112: Why aren't we sending anything to output_port in the Publish messages?

src/consensus/malachite/spawn.rs
```diff
@@ -0,0 +1,197 @@
...
+        shard_validator,
+    };
+    let actor_ref = Host::spawn(state).await?;
+    Ok(actor_ref)
+}
+
+pub async fn spawn_consensus_actor(
+    ctx: SnapchainValidatorContext,
+    timeout_cfg: TimeoutConfig,
+    initial_height: <SnapchainValidatorContext as informalsystems_malachitebft_core_types::Context>::Height,
```
aditiharini, line 66: Can we add some type definitions outside to make these arg types more readable?

PR #237, opened 2025-01-31
allow unlimited mempool size

aditiharini:
We keep dropping messages in the mempool during backfills. Allow unlimited mempool size for backfills.

### DIFF
--- a/src/mempool/mempool.rs
+++ b/src/mempool/mempool.rs
@@ -27,11 +27,15 @@ use tracing::{error, warn};
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct Config {
     pub queue_size: u32,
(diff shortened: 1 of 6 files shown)


src/mempool/mempool.rs
```diff
@@ -27,11 +27,15 @@ use tracing::{error, warn};
...
     pub queue_size: u32,
+    pub allow_unlimited_mempool_size: bool,
 }
 
 impl Default for Config {
     fn default() -> Self {
-        Self { queue_size: 500 }
+        Self {
+            queue_size: 500,
+            allow_unlimited_mempool_size: false,
```
sanjayprabhu, line 37: Typically we just set queue_size to 0 to mean unlimited. But this is fine for now.

sanjayprabhu (approved):

PR #242, opened 2025-02-03
feat: additional read apis for grpc, fleshing out http server

CassOnMars:
Adds the other read apis we can currently support for gRPC, and scaffolds out most of the http api, but isn't exposed via mod so cargo doesn't try to compile it


src/network/server.rs
```diff
@@ -866,6 +886,573 @@ impl HubService for MyHubService {
...
+        if per_shard_tokens.len() != num_shards {
+            return Err(Status::invalid_argument(
+                "Page token does not match number of shards".to_string(),
+            ));
+        }
+
+        let pages: Vec<MessagesPage> =
+            self.shard_stores
+                .iter()
+                .zip(per_shard_tokens.into_iter())
```
sanjayprabhu, line 974: Ah nice, this is a good idea.  I wonder if we should make this explicit by adding a shard id to the request. Probably worth doing sometime after mainnet (and hubs are deprecated).

src/network/server.rs
```diff
@@ -866,6 +886,573 @@ impl HubService for MyHubService {
...
+            return Err(Status::invalid_argument(
+                "Page token does not match number of shards".to_string(),
+            ));
+        }
+
+        let pages: Vec<MessagesPage> = self
+            .shard_stores
+            .iter()
+            .zip(per_shard_tokens.into_iter())
+            .map(|(shard_entry, shard_token)| {
```
sanjayprabhu, line 1127: This is duplicated a lot, can we push this into shard_stores?

sanjayprabhu (approved):

PR #244, opened 2025-02-03
pick up on chain events ingestion where it left off

aditiharini:
Store the last ingested onchain event and pick up ingestion where we left off.


src/proto/ingest_state.proto
```diff
@@ -0,0 +1,5 @@
+syntax = "proto3";
+
+message OnChainEventState {
```
sanjayprabhu, line 3: Let's call the proto file something more generic, like `node_state.proto`.

src/bin/setup_local_testnet.rs
```diff
@@ -142,7 +145,7 @@ num_shards = {num_shards}
 [onchain_events]
 rpc_url= "{l2_rpc_url}"
 start_block_number = {start_block_number}
-stop_block_number = {stop_block_number}
+{stop_block_number}
```
sanjayprabhu, line 148: Let's keep `stop_block_number={stop_block_number}` for consistency

src/connectors/onchain_events/mod.rs
```diff
@@ -249,6 +255,25 @@ impl Subscriber {
             }
             Some(events) => events.push(event.clone()),
         }
+        if block_number as u64 > self.latest_block_in_db() {
+            match ingest_state::onchain_events::put_state(
```
sanjayprabhu, line 259: This is at the wrong level. `add_onchain_event` gets call multiple times per block. This should be at least one level higher, in process_log. But it would be even better to do this one per batch when syncing historical events in bulk, and then once per block when syncing live events.

src/main.rs
```diff
@@ -196,14 +197,21 @@ async fn main() -> Result<(), Box<dyn Error>> {
         });
     }
 
+    let global_db_dir = path::Path::new(&app_config.rocksdb_dir)
+        .join("global")
+        .to_string_lossy()
+        .into_owned();
+    let global_db = Arc::new(RocksDB::new(&global_db_dir));
+    global_db.open().unwrap();
```
sanjayprabhu, line 205: We should push all this into the store.

src/storage/store/engine.rs
```diff
@@ -520,6 +521,13 @@ impl ShardEngine {
                         self.update_trie(trie_ctx, &hub_event, txn_batch)?;
                         events.push(hub_event.clone());
                         system_messages_count += 1;
+                        match max_block_number {
+                            None => max_block_number = Some(onchain_event.block_number),
+                            Some(block_number) => {
+                                max_block_number =
+                                    Some(onchain_event.block_number.max(block_number))
```
sanjayprabhu, line 528: Why do we need this?

src/storage/store/ingest_state.rs
```diff
@@ -0,0 +1,36 @@
...
+#[derive(Error, Debug)]
+pub enum IngestStateError {
+    #[error(transparent)]
+    DecodeError(#[from] DecodeError),
+
+    #[error(transparent)]
+    RocksdbError(#[from] RocksdbError),
+}
+
+pub mod onchain_events {
```
sanjayprabhu, line 13: Let's call this NodeLocalState

src/storage/constants.rs
```diff
@@ -39,6 +38,8 @@ pub enum RootPrefix {
 
     /* Used to index user submitted username proofs */
     UserNameProofByName = 16,
+
+    OnChainEventState = 17,
```
sanjayprabhu, line 42: `NodeLocalState`

src/main.rs
aditiharini: Which store? I was planning to use the same db for both onchain events and fnames.

src/storage/constants.rs
sanjayprabhu: I was thinking those can be postfixes of the same root prefix. But actually, we shouldn't do this at the `RootPrefix` enum. I think its better to do a `LocalStateRootPrefix` because it's going to be in a separate DB and these two prefixes should never collide.

src/main.rs
sanjayprabhu: The file currently called ingest_state

PR #245, opened 2025-02-04
clean up on chain event ingest persistence

aditiharini:
Clean up code for onchain events persistence.


src/connectors/onchain_events/mod.rs
```diff
@@ -524,7 +533,7 @@ impl Subscriber {
     }
 
     fn latest_block_in_db(&self) -> u64 {
-        match ingest_state::onchain_events::get_state(&self.db) {
+        match node_local_state::onchain_events::get_state(&self.db) {
```
sanjayprabhu, line 536: We should make this a stuct, and have it construct the db itself. Since we have so many DBs (per shard, local state etc) if we mix up which db we're passing into the component, we could corrupt data. 

This should look more like:

```
node_local_store.get_latest_on_chain_block_number() -> Result<u64, Err>
node_local_store.set_latest_on_chain_block_number(u64) -> Result<(), Err>
```

etc

sanjayprabhu (dismissed):

PR #246, opened 2025-02-04
pick up fname transfer ingest from where it left off

aditiharini:
Start ingesting fname transfer from where the node stopped before restart.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -186,10 +186,14 @@ async fn main() -> Result<(), Box<dyn Error>> {
         app_config.fc_network,
     );
 
+    let global_db = RocksDB::open_global_db(&app_config.rocksdb_dir);
+    let local_state_store = LocalStateStore::new(global_db);
+
     if !app_config.fnames.disable {
         let mut fetcher = snapchain::connectors::fname::Fetcher::new(
             app_config.fnames.clone(),
             mempool_tx.clone(),
+            local_state_store.clone(),
         );
 
         tokio::spawn(async move {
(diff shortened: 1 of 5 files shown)


src/storage/constants.rs
```diff
@@ -40,6 +40,8 @@ pub enum RootPrefix {
     UserNameProofByName = 16,
 
     OnChainEventState = 17,
+
+    FnameTransferState = 18,
 }
```
sanjayprabhu, line 45: See comment on other PR, we should move this out of RootPrefix

src/storage/store/node_local_state.rs
```diff
@@ -29,7 +29,29 @@ pub mod onchain_events {
...
     pub fn get_state(db: &RocksDB) -> Result<Option<OnChainEventState>, super::IngestStateError> {
         match db.get(&make_primary_key())? {
-            Some(hub_state) => Ok(Some(OnChainEventState::decode(hub_state.as_slice())?)),
+            Some(state) => Ok(Some(OnChainEventState::decode(state.as_slice())?)),
+            None => Ok(None),
+        }
+    }
+}
+
+pub mod fname_transfers {
```
sanjayprabhu, line 38: Let's make this a struct, and make the functions less generic (set_latest_transfer_id/get_latest_transfer_id)

src/main.rs
```diff
@@ -186,23 +186,25 @@ async fn main() -> Result<(), Box<dyn Error>> {
         app_config.fc_network,
     );
 
+    let global_db_dir = path::Path::new(&app_config.rocksdb_dir)
+        .join("global")
+        .to_string_lossy()
+        .into_owned();
+    let global_db = Arc::new(RocksDB::new(&global_db_dir));
+    global_db.open().unwrap();
```
sanjayprabhu, line 194: `RocksDb::open_global_db`

PR #248, opened 2025-02-05
make mempool poll interval configurable

aditiharini:
Right now, the mempool always stays very small. This points to us not reading in messages as quickly as we're reading them out. Make the poll interval configurable so we can play around with how fast we read messages in.

### DIFF
--- a/src/mempool/mempool.rs
+++ b/src/mempool/mempool.rs
@@ -41,7 +41,7 @@ impl Default for Config {
             queue_size: 500,
             allow_unlimited_mempool_size: false,
             capacity_per_shard: 1024,
-            rx_poll_interval: Duration::from_micros(1),
+            rx_poll_interval: Duration::from_millis(1),
         }
     }
 }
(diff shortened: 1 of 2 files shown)


src/mempool/mempool.rs
```diff
@@ -37,6 +41,7 @@ impl Default for Config {
             queue_size: 500,
             allow_unlimited_mempool_size: false,
             capacity_per_shard: 1024,
+            rx_poll_interval: Duration::from_micros(1),
```
sanjayprabhu, line 44: Do we want this to be in microseconds? If so, we should call it `rx_poll_interval_microsecs` so it's clear. I think this is way to aggressive and will cause it's own performance problems. The lowest we should go is 1ms IMO. What's the issue this is trying to fix?

aditiharini: It's specified as a `Duration` and not an int so you have to feed in a string that has "ms", "s", "us" attached to it.

aditiharini: In the logs, I'm seeing that each request to the mempool is only really pulling 1 message out of the mempool. Before, we were seeing 150 pulled at a time (which is the limit). I want the mempool to be fully saturated so that we do make maximum use of the bandwidth available for block production.

PR #250, opened 2025-02-06
bugfixes for onchain events and fname ingest persistence

aditiharini:
Fix onchain event ingest to persist a block number for historical sync only after all event types have been ingested for that block number. Fix fname ingest to persist the latest username proof rather than block numbers.

### DIFF
--- a/src/connectors/onchain_events/mod.rs
+++ b/src/connectors/onchain_events/mod.rs
@@ -270,24 +270,19 @@ impl Subscriber {
         }
     }
 
-    fn record_block_number(&self, event: &Log) {
-        match event.block_number {
-            Some(block_number) => {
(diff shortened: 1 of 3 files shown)


src/connectors/onchain_events/mod.rs
```diff
@@ -478,40 +473,50 @@ impl Subscriber {
...
+        start_block: u64,
+        stop_block: u64,
+    ) -> Result<(), SubscribeError> {
+        let filter = Filter::new()
+            .address(address)
+            .from_block(start_block)
+            .to_block(stop_block);
+        info!(
+            start_block,
+            stop_block, "Syncing historical events in range"
```
sanjayprabhu, line 488: Should include the event type here

sanjayprabhu (approved):

PR #253, opened 2025-02-06
observability and fixes for ingest to mempool communication

aditiharini:
Add metrics for onchain events and fname ingest. Also lumped in a couple small fixes in these areas + communication with mempool.

### DIFF
--- a/src/connectors/fname/mod.rs
+++ b/src/connectors/fname/mod.rs
@@ -9,6 +9,7 @@ use tracing::{debug, error, info, warn};
 use crate::{
     proto::{FnameTransfer, UserNameProof, UserNameType, ValidatorMessage},
     storage::store::{engine::MempoolMessage, node_local_state::LocalStateStore},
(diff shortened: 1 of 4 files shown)


src/connectors/fname/mod.rs
```diff
@@ -161,6 +167,7 @@ impl Fetcher {
                     fid: t.to,
                     r#type: UserNameType::UsernameTypeFname as i32,
                 };
+                self.count("num_fname_transfers_ingested", 1);
```
sanjayprabhu, line 170: We keep metric names concise by convention. So something like `fnames.transfers`

src/connectors/fname/mod.rs
```diff
@@ -125,6 +127,11 @@ impl Fetcher {
         }
     }
 
+    fn count(&self, key: &str, value: u64) {
```
sanjayprabhu, line 130: Might also be worth adding a guage for the highest transfer id we've seen

sanjayprabhu (approved):

PR #255, opened 2025-02-07
debug consensus

aditiharini:


src/consensus/malachite/host.rs
```diff
@@ -197,7 +194,6 @@ impl Host {
                     }
                 };
                 let proposed_value = state.shard_validator.add_proposed_value(&proposal);
-                state.shard_validator.decide(commits).await;
```
sanjayprabhu, line 200: We should also add error or panic within `Proposer.decide` if we can't find the proposal to commit

PR #257, opened 2025-02-13
implemented a fixed-size queue for the mempool

cryptoryda:
I implemented a fixed-size queue for the mempool using `BTreeMap` to prevent unbounded memory growth. A `MAX_MEMPOOL_SIZE` limit (10,000 messages per shard) is enforced, with an eviction strategy that removes the oldest message when the limit is reached!


sanjayprabhu:
Already implemented.

PR #258, opened 2025-02-13
 caching mechanism in the Subscriber 

cryptoryda:
I added a caching mechanism in the Subscriber by introducing a `timestamp_cache` field (a `Mutex<HashMap<FixedBytes<32>, u64>>`) and modified the `get_block_timestamp` method to check this cache before making an RPC call. This reduces redundant network  requests for block timestamps, improving performance and reducing overhead, especially under high-load conditions.

### DIFF
--- a/src/connectors/onchain_events/mod.rs
+++ b/src/connectors/onchain_events/mod.rs
@@ -12,6 +12,7 @@ use serde::{Deserialize, Serialize};
 use thiserror::Error;
 use tokio::sync::mpsc;
 use tracing::{error, info};
+use tokio::sync::Mutex;
 
 use crate::{
     proto::{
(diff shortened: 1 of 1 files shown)


sanjayprabhu:
Thanks, this is useful. Please fix conflicts and we'll merge.

src/connectors/onchain_events/mod.rs
```diff
@@ -146,6 +147,7 @@ pub struct Subscriber {
     start_block_number: u64,
     stop_block_number: u64,
     statsd_client: StatsdClientWrapper,
+    timestamp_cache: Mutex<HashMap<FixedBytes<32>, u64>>,
```
aditiharini, line 150: Why is this wrapped in a mutex?

src/connectors/onchain_events/mod.rs
```diff
@@ -167,6 +169,7 @@ impl Subscriber {
             start_block_number: config.start_block_number,
             stop_block_number: config.stop_block_number,
             statsd_client,
+            timestamp_cache: Mutex::new(HashMap::new()),
```
aditiharini, line 172: The cache should have some max capacity so we don't leak memory. You can use the moka library for an lru cache.

PR #259, opened 2025-02-14
fix: parse fname server response correctly

CassOnMars:


src/storage/store/engine.rs
```diff
@@ -536,61 +536,76 @@ impl ShardEngine {
...
-                    Err(err) => {
-                        warn!("Error merging fname transfer: {:?}", err);
+                match &fname_transfer.proof {
+                    None => {
+                        warn!(
+                            fid = snapchain_txn.fid,
+                            id = fname_transfer.id,
+                            "Fname transfer has no proof"
+                        );
+                        return Err(EngineError::EngineMessageValidationError(
```
sanjayprabhu, line 546: I don't think we should return here because that would skip processing the rest of the messages in the transaction. 

We should probably break up this function into smaller functions so it's easier to reason about and we can't make this kind of mistake. I'll do this in a separate PR.

src/storage/store/engine_tests.rs
```diff
@@ -1236,6 +1219,8 @@ mod tests {
         commit_message(&mut engine, &username_add).await;
     }
 
+    // this test needs to be updated to use an actual transfer event since validation logic checks the fname signer signature
+    #[ignore]
```
sanjayprabhu, line 1223: Hmm, we should actually pull the signer address from the validations from a config, so tests can pass in a custom public key.

src/storage/store/engine_tests.rs
```diff
@@ -1170,33 +1179,7 @@ mod tests {
...
-            transfer_event.r#type,
-            proto::HubEventType::MergeUsernameProof as i32
-        );
-
-        // don't insert an fname for fid 0 into the trie
-        assert!(!test_helper::key_exists_in_trie(
-            &mut engine,
-            &TrieKey::for_fname(0, fname)
-        ));
+        // todo: alternate signer for fname server testing so we can verify transferring to 0 nukes the fname
```
sanjayprabhu, line 1182: This is hard to do because of the signature validation? See note below. I think `username_factory::create_transfer` should use custom hardcoded signer for tests, and the engine should pull in the public key to use for validations from config.

sanjayprabhu (approved):

PR #260, opened 2025-02-14
don't call decide on synced values

aditiharini:
We currently try to commit synced values twice because malachite calls `decide` internally and we call `decide` on seeing a synced value.

### DIFF
--- a/src/consensus/proposer.rs
+++ b/src/consensus/proposer.rs
@@ -183,6 +183,13 @@ impl Proposer for ShardProposer {
             self.publish_new_shard_chunk(&chunk.clone()).await;
             self.engine.commit_shard_chunk(&chunk);
             self.proposed_chunks.remove(&value);
+        } else {
+            error!(
+                height = commits.height.unwrap().to_string(),
+                round = commits.round,
+                shard_hash = hex::encode(value.hash),
+                "Unable to find proposal for decided value"
+            )
         }
         self.statsd_client.gauge_with_shard(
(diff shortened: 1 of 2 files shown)


src/consensus/proposer.rs
```diff
@@ -183,6 +183,13 @@ impl Proposer for ShardProposer {
             self.publish_new_shard_chunk(&chunk.clone()).await;
             self.engine.commit_shard_chunk(&chunk);
             self.proposed_chunks.remove(&value);
+        } else {
+            error!(
+                height = commits.height.unwrap().to_string(),
+                round = commits.round,
+                shard_hash = hex::encode(value.hash),
+                "Unable to find proposal for decided value"
```
sanjayprabhu, line 191: I think we should panic here. We cannot progress after this, because this node will not be able to propose or validate any subsequent height correctly. It would not have committed a chunk to disk.

src/consensus/proposer.rs
```diff
@@ -426,6 +433,13 @@ impl Proposer for BlockProposer {
             self.engine.commit_block(block);
             self.proposed_blocks.remove(&value);
             self.pending_chunks.remove(&height.block_number);
+        } else {
+            error!(
+                height = height.to_string(),
+                round = commits.round,
+                shard_hash = hex::encode(&commits.value.unwrap().hash),
+                "Unable to find proposal for decided value"
+            )
```
sanjayprabhu, line 442: Same here.

PR #261, opened 2025-02-14
store undecided proposals

aditiharini:
In order to recover from restarts gracefully, we need to store undecided proposals and make sure to propose the same value for each height/round even across restarts.


src/storage/store/proposal.rs
```diff
@@ -0,0 +1,97 @@
...
+    DecodeError(#[from] prost::DecodeError),
+}
+
+/** A page of messages returned from various APIs */
+pub struct BlockPage {
+    pub blocks: Vec<Block>,
+    pub next_page_token: Option<Vec<u8>>,
+}
+
+fn make_proposal_key(height: u64, round: i64) -> Vec<u8> {
```
sanjayprabhu, line 34: We should add the shard index (as the first element) to be safe.

src/storage/store/proposal.rs
```diff
@@ -0,0 +1,97 @@
...
+    // Store the prefix in the first byte so there's no overlap across different stores
+    let mut key = vec![RootPrefix::Block as u8];
+    // Store the block number in the next 8 bytes
+    key.extend_from_slice(&height.to_be_bytes());
+    key.extend_from_slice(&round.to_be_bytes());
+
+    key
+}
+
+pub fn put_proposal(db: &RocksDB, proposal: FullProposal) -> Result<(), ProposalStorageError> {
```
sanjayprabhu, line 44: These functions can go into the structs impls, no need to keep them separate. The hub used to do this, but I think it's not as clean. We should start fixing it.

src/consensus/validator.rs
```diff
@@ -131,6 +135,12 @@ impl ShardValidator {
         } else {
             panic!("No proposer set");
         }
+        if let Err(err) = self
+            .proposal_store
+            .delete_proposal(commits.height.unwrap(), Round::from(commits.round))
```
sanjayprabhu, line 140: This should delete for every round for the height. Because it's possible the proposal proposed by this node at a particular round was actually passed on, and another node proposed a different value at a higher round.

src/storage/constants.rs
```diff
@@ -41,6 +41,8 @@ pub enum RootPrefix {
 
     /* Used to maintain information about latest onchain events, fnames ingested */
     NodeLocalState = 17,
+
+    Proposal = 18,
```
sanjayprabhu, line 45: Hmm, I wonder if this should go under NodeLocalState (and use the node local db instead of the shard dbs). Mainly because this is specific to the proposer. e.g. You don't want this info included in the snapshot. It doesn't hurt because it's unlikely it will matter when resuming from snapshot, but it's cleaner to not have to worry about this kind of issue in the future.

src/consensus/validator.rs
```diff
@@ -187,12 +201,25 @@ impl ShardValidator {
         round: Round,
         timeout: Duration,
     ) -> FullProposal {
-        if let Some(block_proposer) = &mut self.block_proposer {
+        match self.proposal_store.get_proposal(height, round) {
```
sanjayprabhu, line 204: It's a bit annoying we have to do this for every propose. I think we only need to do this the first time we restart? But not a huge deal, we can optimize later.

src/consensus/validator.rs
```diff
@@ -168,6 +178,10 @@ impl ShardValidator {
             panic!("No proposer set");
         };
 
+        if let Err(err) = self.proposal_store.put_proposal(full_proposal.clone()) {
+            error!("Unable to store proposal {}", err.to_string());
+        };
```
sanjayprabhu, line 183: I don't think we should do this in `add_proposed_value`, that's going to persist it for every node (and during sync). We only need to remember our own proposals.

src/storage/constants.rs
aditiharini: Yeah, I think it makes sense to put in `NodeLocalState`. Moved.

src/consensus/validator.rs
aditiharini: Yeah, agreed. Will leave as-is for now and we can optimize later.

PR #264, opened 2025-02-15
fix: Block production fixes

sanjayprabhu:
- Add consistent 250ms (by default) block time via commit timeout
- Don't store full shard chunks in the block, just store headers
- Look up chunks from shard db instead of waiting on the channel (fixes block proposer getting stuck forever if it didn't get a chunk in time)
- Fix consensus tests not correctly including messages
- Delete unused sync code
- set max_messages per block to 500 to compensate for slower block time (~1mbs at 200 bytes per messages * 500 * 2 shards * 4 blocks per sec at 250ms). We probably have some room to bump this higher, but need to be mindful of the number of shards running on the same node.


src/consensus/malachite/spawn.rs
```diff
@@ -181,9 +182,12 @@ impl MalachiteConsensusActors {
         )
         .await?;
 
+        let mut timeout_config = TimeoutConfig::default();
+        // setting the commit timeout sets up a fixed block production rate.
```
sanjayprabhu, line 186: This affects and slows down sync right now, but malachite has already fixed this. So we should get it when we upgrade next week.

src/consensus/proposer.rs
```diff
@@ -297,56 +311,69 @@ impl BlockProposer {
...
-                            self.pending_chunks.insert(chunk_block_number, vec![chunk]);
+                    for (shard_id, store) in self.shard_stores.iter() {
+                        if chunks.contains_key(shard_id) {
+                            continue;
                         }
-                    }
-                    if let Some(chunks) = self.pending_chunks.get(&requested_height) {
-                        if chunks.len() == self.num_shards as usize {
-                            break;
+                        let result = store.shard_store.get_chunk_by_height(requested_height);
```
sanjayprabhu, line 343: Switches us to look in the db instead of waiting for the channel which was causing halts (if we lose a message, we could never recover).

We'll need to refactor once we want to allow block validators to run separately from shard validators, but that's going to take a while.

src/consensus/proposer.rs
```diff
@@ -415,9 +447,58 @@ impl Proposer for BlockProposer {
...
     fn add_proposed_value(&mut self, full_proposal: &FullProposal) -> Validity {
-        if let Some(proto::full_proposal::ProposedValue::Block(_block)) =
+        if let Some(proto::full_proposal::ProposedValue::Block(block)) =
             &full_proposal.proposed_value
         {
+            let header = block.header.as_ref().unwrap();
+            let height = header.height.unwrap();
+
+            if height != self.get_confirmed_height().increment() {
+                error!(
```
sanjayprabhu, line 457: This fixes a lot of the merkle trie root mismatch issues, because for some reason we were validating the message after deciding on it:
```
2025-02-15T01:48:30.196712Z ERROR snapchain::consensus::proposer: Cannot validate height, not the next height shard=1 our_height=13 proposal_height=13
2025-02-15T01:48:30.217653Z ERROR snapchain::consensus::proposer: Cannot validate height, not the next height shard=0 our_height=12 proposal_height=12
2025-02-15T01:48:30.569546Z ERROR snapchain::consensus::proposer: Cannot validate height, not the next height shard=1 our_height=14 proposal_height=14
```

Will need to figure out why, but this makes the logs less noisy and scary when it does happen.

src/consensus/proposer.rs
```diff
@@ -415,9 +447,58 @@ impl Proposer for BlockProposer {
...
+            if header.shard_witnesses_hash.is_empty() {
+                error!("Received block with missing shard witnesses hash");
+                return Validity::Invalid;
+            }
+            if block.shard_witness.is_none() {
+                error!("Received block with missing shard witnesses");
+                return Validity::Invalid;
+            }
+            let witness = block.shard_witness.as_ref().unwrap();
+            if witness.shard_chunk_witnesses.len() != self.num_shards as usize {
```
sanjayprabhu, line 490: We need unit tests for the proposer now, let me look into it. I might just do it in a separate pr.

tests/consensus_test.rs
```diff
@@ -231,15 +236,15 @@ impl NodeForTest {
...
     pub async fn total_messages(&self) -> usize {
-        let messages = self
-            .block_store
-            .get_blocks(0, None, &PageOptions::default())
-            .unwrap()
-            .blocks
-            .into_iter()
-            .map(|b| b.shard_chunks[0].transactions[0].user_messages.len());
-
-        messages.len()
```
sanjayprabhu, line 242: This was not actually counting the messages. And no messages were included in the blocks because the events didn't exist.

src/consensus/malachite/spawn.rs
```diff
@@ -181,9 +182,12 @@ impl MalachiteConsensusActors {
...
+        timeout_config.timeout_commit = config.propose_value_delay;
```
aditiharini, line 187: We already use `propose_value_delay` elsewhere. Should we create a new config parameter for the existing use case? It seems like the 2 places where we use this shouldn't share the same config.

src/proto/blocks.proto
```diff
@@ -92,19 +92,26 @@ message BlockHeader {
...
+  Height height = 1;
+  bytes shard_root = 2;
+  bytes shard_hash = 3;
+}
 
 message Block {
   BlockHeader header = 1;
   bytes hash = 2;
-  repeated ShardChunk shard_chunks = 3;
-  optional ValidatorSet validators = 4;
```
aditiharini, line 106: Why did we need to remove the validator set?

src/consensus/malachite/spawn.rs
sanjayprabhu: Fixed.

src/consensus/proposer.rs
sanjayprabhu: Looks like this is happening because `add_proposed_value` is being called multiple times. We should handle this better, but keeping x most recently decided blocks around in memory. Otherwise we'll be voting invalid on something we previously voted as valid. I'll ticket this separately.

src/consensus/consensus.rs
```diff
@@ -75,8 +75,8 @@ impl Default for Config {
             private_key: hex::encode(SecretKey::generate()),
             shard_ids: vec![1],
             num_shards: 1,
-            propose_value_delay: Duration::from_millis(250),
-            max_messages_per_block: 250, //TODO
+            block_time: Duration::from_millis(250),
+            max_messages_per_block: 500,
```
sanjayprabhu, line 79: ~1mbs at 200 bytes per messages * 500 * 2 shards * 4 blocks per sec at 250ms. 

We probably have some room to bump this higher, but need to be mindful of the number of shards running on the same node.

PR #267, opened 2025-02-20
support read nodes

aditiharini:
Add support for nodes that can listen to committed blocks and serve rpcs but won't participate in consensus.


src/consensus/malachite/host.rs
```diff
@@ -161,11 +164,31 @@ impl Host {
...
+                } else {
+                    None
+                };
+
                 state
-                    .shard_validator
-                    .decide(Commits::from_commit_certificate(&certificate))
-                    .await;
+                    .gossip_tx
+                    .send(GossipEvent::BroadcastDecidedValue(proto::DecidedValue {
```
sanjayprabhu, line 186: Can we not re-use the shard decision tx/rx channels that the mempool uses?

src/consensus/malachite/read_host.rs
```diff
@@ -0,0 +1,146 @@
...
+        _myself: ActorRef<ReadHostMsg>,
+        msg: ReadHostMsg,
+        state: &mut ReadHostState,
+    ) -> Result<(), ActorProcessingErr> {
+        match msg {
+            ReadHostMsg::GetHistoryMinHeight { reply_to } => {
+                reply_to.send(crate::proto::Height::new(state.shard_id, 1))?;
+            }
+
+            ReadHostMsg::ProcessDecidedValue { value } => match &mut state.engine {
```
sanjayprabhu, line 69: I think we'll need to have some state here to handle missing heights and non-consecutive heights. The ordering of the messages here is not guaranteed. 

Also, we should add an explicit check within `commit_...` to check that we're replaying the very next height not not anything else (maybe within `replay_proposal`?)

src/consensus/malachite/spawn.rs
```diff
@@ -127,6 +142,96 @@ pub async fn spawn_sync_actor(
...
+
+    let metrics = SyncMetrics::register(registry);
+
+    let actor_ref = ReadSync::spawn(ctx, network, host, params, metrics, span).await?;
+
+    Ok(actor_ref)
+}
+
+#[derive(Clone)]
+pub struct MalachiteReadNodeActors {
```
sanjayprabhu, line 166: we might want to split all the read node code into a separate file

src/consensus/consensus.rs
```diff
@@ -17,6 +18,8 @@ pub enum MalachiteEventShard {
 pub enum SystemMessage {
     MalachiteNetwork(MalachiteEventShard, MalachiteNetworkEvent), // Shard Id and the malachite network event
     Mempool(MempoolMessageWithSource),
+
+    DecidedValue(proto::DecidedValue),
```
sanjayprabhu, line 22: We should make this clearer that it's for a read node only, maybe something like `ReadNodeReceiveDecidedValue`

src/consensus/read_validator.rs
```diff
@@ -0,0 +1,145 @@
...
+                    panic!("Invalid decided value")
+                }
+            },
+        };
+        self.last_height = height;
+        sync.cast(read_sync::Msg::Decided(height)).unwrap();
+    }
+
+    fn process_buffered_blocks(&mut self, sync: &read_sync::ReadSyncRef) {
+        while let Some((height, value)) = self.buffered_blocks.pop_first() {
```
sanjayprabhu, line 74: This is guaranteed to be sorted by height?

src/main.rs
```diff
@@ -154,204 +159,254 @@ async fn main() -> Result<(), Box<dyn Error>> {
     // Use the new non-global metrics registry when we upgrade to newer version of malachite
     let _ = Metrics::register(registry);
     let (messages_request_tx, messages_request_rx) = mpsc::channel(100);
-    let (shard_decision_tx, shard_decision_rx) = broadcast::channel(100);
 
-    let global_db = RocksDB::open_global_db(&app_config.rocksdb_dir);
-    let local_state_store = LocalStateStore::new(global_db);
+    if app_config.read_node {
```
sanjayprabhu, line 163: main is getting rather large, and now with two distinct branches it's harder to reason about. Worth splitting it into two, but can happen in a followup

src/proto/blocks.proto
```diff
@@ -79,6 +79,19 @@ message FullProposal {
...
 
+message DecidedValue {
+  oneof value {
+    Block block = 2;
+    ShardChunk shard = 3;
+  }
+}
+
+message ReadNodeMessage {
+  oneof read_node_message {
```
sanjayprabhu, line 90: Why do we need a oneof here?

src/storage/store/engine.rs
```diff
@@ -148,6 +148,7 @@ struct TransactionCounts {
     system_messages: u64,
 }
 
+#[derive(Clone)]
 pub struct ShardEngine {
```
sanjayprabhu, line 152: Hmm, I'm not entirely sure if this is safe to clone, there might be some weirdness with the trie in-memory state. I think it's better to construct a new instance. This is what we're doing with the SnapchainNode.

src/consensus/malachite/read_host.rs
```diff
@@ -0,0 +1,111 @@
...
+                state.validator.initialize_height(&sync);
+            }
+
+            ReadHostMsg::GetHistoryMinHeight { reply_to } => {
+                // For now until we implement pruning
+                reply_to.send(state.validator.get_min_height())?;
+            }
+
+            ReadHostMsg::ProcessDecidedValue { value, sync } => {
+                state.validator.process_decided_value(value, &sync);
```
sanjayprabhu, line 73: Same here, don't pass `sync` to the validator

src/proto/blocks.proto
aditiharini: This is in case we add other messages specific to read nodes. We can add something without a breaking protocol change.

.github/workflows/verify.yml
```diff
@@ -43,6 +43,13 @@ jobs:
           ref: ${{ env.MALACHITE_GIT_REF }}
           path: ./malachite
 
+      - name: Install Rust 1.85.0
+        uses: actions-rs/toolchain@v1
+        with:
+          toolchain: 1.85.0
```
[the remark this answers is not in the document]
aditiharini: The runners apparently have some rust version installed and it defaults to that.

src/consensus/malachite/host.rs
aditiharini: Sounds good, I'm gonna defer this because though the shard_decision_rx is available here, it requires extra work to handle it in the gossip code.

src/consensus/read_validator.rs
```diff
@@ -0,0 +1,145 @@
...
+            if (self.buffered_blocks.len() as u32) < self.max_num_buffered_blocks {
+                self.buffered_blocks.insert(height, value);
+            } else {
+                info!(%height, last_height = %self.last_height, "Dropping decided block because buffered block space is full")
+            }
+        } else if height == self.last_height.increment() {
+            self.commit_decided_value(&value, height, &sync);
+            self.process_buffered_blocks(&sync);
+        } else {
+            info!(%height, last_height = %self.last_height, "Dropping decided block because height is too low")
```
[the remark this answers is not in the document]
aditiharini: Why? This happens because we broadcast decided blocks from all validators. All except the first one hits this log. Maybe worth downgrading to debug but would not upgrade to warn.

src/main.rs
aditiharini: Yeah, agreed that this is messy. Let's address in a follow up. There are significant changes here for supporting rpcs + mempool so would be good to evaluate what to do after those changes.

src/storage/store/engine.rs
aditiharini: Done, this was only used in tests but I worked around it. Just want to pass the same db instance around.

src/consensus/read_validator.rs
aditiharini: Yeah, the BTreeMap is sorted by key. I added a comment to make it clear that this is important and prompt us to think before changing.

src/consensus/malachite/read_sync.rs
```diff
@@ -0,0 +1,455 @@
...
+                        inflight.insert(request_id.clone(), InflightRequest { peer_id, request });
+                    }
+                    Err(e) => {
+                        error!("Failed to send request to gossip layer: {e}");
+                    }
+                }
+            }
+
+            Effect::SendValueResponse(request_id, value_response) => {
+                warn!(
```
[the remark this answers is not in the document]
aditiharini: Downgraded a lot of logs from info to debug.

src/consensus/read_validator.rs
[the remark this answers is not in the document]
sanjayprabhu: Ah makes sense. I misread it. Agree this should be debug or we shouldn't log at all

src/consensus/malachite/read_host.rs
```diff
@@ -70,7 +72,11 @@ impl ReadHost {
             }
 
             ReadHostMsg::ProcessDecidedValue { value, sync } => {
-                state.validator.process_decided_value(value, &sync);
+                let num_values_processed = state.validator.process_decided_value(value);
+                if num_values_processed > 0 {
+                    sync.cast(read_sync::Msg::Decided(state.validator.last_height))
+                        .unwrap();
```
sanjayprabhu, line 78: These unwraps are fine since it's just a read node, but we have to be careful about panics. It's better to use `?` and bubble up the error

src/consensus/read_validator.rs
```diff
@@ -93,20 +96,35 @@ impl ReadValidator {
...
+            0
+        };
+        self.statsd_client.gauge_with_shard(
+            self.shard_id,
+            "read_validator.num_buffered_blocks",
+            self.buffered_blocks.len() as u64,
+        );
+        self.statsd_client.count_with_shard(
+            self.shard_id,
+            "read_validator.num_commited_values",
```
sanjayprabhu, line 124: This is just block height right? Do we need to duplicate it?

sanjayprabhu (approved):

src/consensus/read_validator.rs
aditiharini: We don't strictly need it. It would be useful to know if we're receiving the decided values though since it's possible we drop them somewhere before they hit the engine.

src/consensus/malachite/read_host.rs
aditiharini: Fixed

PR #271, opened 2025-02-27
enable rpc server and http server for read nodes

aditiharini:
Tested that rpcs to read nodes work and submitting messages to read nodes works in cluster 2. 
```
 grpcurl -plaintext -proto src/proto/rpc.proto -import-path src/proto 10.0.0.126:13383 HubService/GetInfo
{
  "dbStats": {
    "numMessages": "234569",
    "numFidRegistrations": "1623",
    "approxSize": "138420435"
  },
  "numShards": 2,
  "shardInfos": [
    {
      "maxHeight": "199",
      "approxSize": "174681"
    },
    {
      "shardId": 2,
      "maxHeight": "417",
      "numMessages": "117152",
      "numFidRegistrations": "810",
      "approxSize": "69121768"
    },
    {
      "shardId": 1,
      "maxHeight": "417",
      "numMessages": "117417",
      "numFidRegistrations": "813",
      "approxSize": "69298667"
    }
  ]
}
```

### DIFF
--- a/src/mempool/mempool.rs
+++ b/src/mempool/mempool.rs
@@ -122,6 +122,7 @@ pub struct ReadNodeMempool {
     num_shards: u32,
     mempool_rx: mpsc::Receiver<MempoolMessageWithSource>,
     gossip_tx: mpsc::Sender<GossipEvent<SnapchainValidatorContext>>,
+    statsd_client: StatsdClientWrapper,
 }
 
 impl ReadNodeMempool {
@@ -130,6 +131,7 @@ impl ReadNodeMempool {
         num_shards: u32,
         shard_stores: HashMap<u32, Stores>,
         gossip_tx: mpsc::Sender<GossipEvent<SnapchainValidatorContext>>,
+        statsd_client: StatsdClientWrapper,
     ) -> Self {
         ReadNodeMempool {
             shard_stores,
@@ -137,6 +139,7 @@ impl ReadNodeMempool {
             mempool_rx,
             message_router: Box::new(ShardRouter {}),
             gossip_tx,
+            statsd_client,
         }
     }
 
@@ -239,8 +242,12 @@ impl ReadNodeMempool {
 
     pub async fn run(&mut self) {
         while let Some((message, source)) = self.mempool_rx.recv().await {
+            self.statsd_client
+                .count("read_mempool.messages_received", 1);
             if self.message_is_valid(&message) {
-                self.gossip_message(message, source).await
+                self.gossip_message(message, source).await;
+                self.statsd_client
+                    .count("read_mempool.messages_published", 1);
             }
         }
         panic!("Mempool has exited");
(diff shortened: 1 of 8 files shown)


src/mempool/mempool.rs
```diff
@@ -116,6 +116,139 @@ pub struct MempoolMessagesRequest {
...
+                        }
+                    }
+                }
+                return false;
+            }
+        },
+    }
+}
+
+async fn gossip_message(
```
sanjayprabhu, line 189: I think generally it's better to favor composition (esp. in rust) over small context-less functions. This is small, so not a big deal. 

You could do something like a ReadOnlyMempool which has the validation and gossip features, and then have the ValidatorMempool hold a reference to it and delegate to it where required. Or create a separate struct that holds the common logic that's used by both the ReadNodeMempool and the ValidatorMempool

There are going to be other checks we'll need to add down the line like rate limits and validation (I think we need to do the validation right now as well).

src/mempool/mempool.rs
```diff
@@ -116,6 +116,139 @@ pub struct MempoolMessagesRequest {
...
+        ReadNodeMempool {
+            shard_stores,
+            num_shards,
+            mempool_rx,
+            message_router: Box::new(ShardRouter {}),
+            gossip_tx,
+        }
+    }
+
+    pub fn message_is_valid(&mut self, message: &MempoolMessage) -> bool {
```
sanjayprabhu, line 235: Where is this called?

I think this also needs to do the validations right?

src/mempool/mempool.rs
```diff
@@ -234,6 +212,105 @@ impl Mempool {
...
+    }
+
+    pub async fn run(&mut self) {
+        while let Some((message, source)) = self.mempool_rx.recv().await {
+            self.statsd_client
+                .count("read_mempool.messages_received", 1);
+            if self.message_is_valid(&message) {
+                self.gossip_message(message, source).await;
+                self.statsd_client
+                    .count("read_mempool.messages_published", 1);
```
sanjayprabhu, line 250: Since this is shared, we should remove `read_` from the stats

src/mempool/mempool.rs
```diff
@@ -354,7 +419,7 @@ impl Mempool {
                     // We want to pull in multiple messages per poll so that throughput is not blocked on the polling frequency. The number of messages we pull should be fixed and relatively small so that the mempool isn't always stuck here.
                     for _ in 0..256 {
                         if self.config.allow_unlimited_mempool_size || (self.messages.len() as u64) < self.config.capacity_per_shard {
-                            match self.mempool_rx.try_recv() {
+                            match self.read_node_mempool.mempool_rx.try_recv() {
```
sanjayprabhu, line 422: Hmm, this is looking too deeply into the struct. I know I said composition, but I think look at this, maybe a trait maybe a better way to share the code here. e.g. a `Mempool` trait that implements the validation/gossip etc. 

But this is not a blocking, we can refactor later.

sanjayprabhu (approved):

PR #278, opened 2025-03-01
read nodes subscribe to decided blocks only after initial sync is done

aditiharini:
New blocks can't be processed until earlier ones are synced. Don't start processing new blocks until we're almost done syncing historical blocks.

### DIFF
--- a/src/consensus/malachite/read_sync.rs
+++ b/src/consensus/malachite/read_sync.rs
@@ -114,6 +114,10 @@ impl State {
             return false;
         }
 
+        if self.sync.peers.len() == 0 {
+            return false;
+        }
+
         let sync_height = self.sync.sync_height;
         let peer_ahead_by_threshold = self
             .sync
--- a/src/consensus/malachite/read_sync.rs
+++ b/src/consensus/malachite/read_sync.rs
@@ -104,6 +104,24 @@ pub struct State {
 
     /// Task for sending status updates
     ticker: JoinHandle<()>,
+
+    initial_sync_completed: bool,
+}
+
+impl State {
+    fn initial_sync_first_completed(&mut self) -> bool {
+        if self.initial_sync_completed {
+            return false;
+        }
+
+        let sync_height = self.sync.sync_height;
+        let peer_ahead_by_threshold = self
+            .sync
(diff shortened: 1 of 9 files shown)


src/network/gossip.rs
```diff
@@ -83,6 +83,7 @@ pub enum GossipEvent<Ctx: SnapchainContext> {
     ),
     SyncReply(InboundRequestId, sync::Response<SnapchainValidatorContext>),
     BroadcastDecidedValue(proto::DecidedValue),
+    ReadNodeSynced(),
```
sanjayprabhu, line 86: This breaks the abstraction. This level shouldn't care about app level logic. This should be an instruction to the gossip layer, like `SubscribeToDecidedValuesTopic`

src/network/gossip.rs
```diff
@@ -188,29 +189,20 @@ impl SnapchainGossip {
...
-            // Create a Gossipsub topic
-            let topic = gossipsub::IdentTopic::new(CONSENSUS_TOPIC);
-            // subscribes to our topic
-            let result = swarm.behaviour_mut().gossipsub.subscribe(&topic);
-            if let Err(e) = result {
-                warn!("Failed to subscribe to topic: {:?}", e);
-                return Err(Box::new(e));
-            }
+        // Create a Gossipsub topic
+        let topic = gossipsub::IdentTopic::new(CONSENSUS_TOPIC);
```
sanjayprabhu, line 193: I think we still want the if check? The read nodes shouldn't subscribe to the consensus topic

src/consensus/malachite/read_sync.rs
```diff
@@ -104,6 +104,24 @@ pub struct State {
...
+impl State {
+    fn initial_sync_first_completed(&mut self) -> bool {
+        if self.initial_sync_completed {
+            return false;
+        }
+
+        let sync_height = self.sync.sync_height;
+        let peer_ahead_by_threshold = self
+            .sync
+            .random_peer_with_value(sync_height.increment_by(10));
```
sanjayprabhu, line 120: I don't think we should use a random peer, we should use a validator peer, or we should pick the peer with the highest height (since we know height can never regress)

sanjayprabhu (approved):

PR #279, opened 2025-03-01
feat: Configurable validator sets

FrederikBolding:


src/consensus/validator.rs
```diff
@@ -53,12 +55,33 @@ impl ShardValidator {
...
+        let mut result = &self.validator_set[0];
+        for config in &self.validator_set {
+            if config.effective_at <= height && config.effective_at > result.effective_at {
+                result = config;
+            }
+        }
+        result.clone()
+    }
+
+    pub fn get_validator_set(&self, height: u64) -> SnapchainValidatorSet {
```
sanjayprabhu, line 68: This will be called frequently, maybe worth caching this. Maybe within the config?

sanjayprabhu: Or worth creating a new struct called ValidatorSetConfig that can convert from the config and return the cached SnapchainValidatorSet

src/consensus/consensus.rs
```diff
@@ -22,6 +21,13 @@ pub enum SystemMessage {
     DecidedValueForReadNode(proto::DecidedValue),
 }
 
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct ValidatorSetConfig {
+    pub effective_at: u64,
+    pub validator_public_keys: Vec<String>,
+    pub shard_ids: Vec<u32>,
```
sanjayprabhu, line 28: Could you update `setup_local_testnet` with the new format?

FrederikBolding, line 28: What is the goal with configuring these? Should the validator set only be deemed fit for the shard indices in the array?

src/consensus/validator.rs
FrederikBolding: I created a struct to store this in, but we don't get around cloning the value easily 🤔

src/consensus/consensus.rs
FrederikBolding: Is it meant to be a replacement for the existing `shard_ids` configuration?

FrederikBolding: Done!

sanjayprabhu: Yes, it should only be valid for the shard ids provided. Right now we have the same validator sets for all shards, but it should be possible to have completely separate validator sets for each shard. And change them over time (at specific heights)

src/consensus/consensus.rs
```diff
@@ -32,7 +38,7 @@ pub struct Config {
     pub block_time: Duration,
 
     pub max_messages_per_block: u32,
-    pub validator_addresses: Vec<String>,
```
sanjayprabhu, line 35: Let's keep the existing config for backwards compatibility. I think we'd need to bring back the `validator_set_for` method below, and have it check if validator_addresses is present first and construct ValidatorSetConfig from it.

src/consensus/validator.rs
```diff
@@ -53,12 +82,18 @@ impl ShardValidator {
         }
     }
 
-    pub fn get_validator_set(&self) -> SnapchainValidatorSet {
-        self.validator_set.clone()
+    pub fn get_validator_set(&self, height: u64) -> SnapchainValidatorSet {
+        let mut result = &self.validator_set[0];
```
sanjayprabhu, line 86: We might need to panic somewhere above if validator set length is 0. Preferably as early as possible (in the config, within validator_set_for?)

sanjayprabhu (commented):
Generally looks good, see note about backwards compatibility for existing config files.

src/consensus/consensus.rs
FrederikBolding: Hmm, that seems like it would complicate things a lot, would it not? Then both of the options have to be optional and we would need handling for if one or the other or both are defined? What takes precedence if both are specified for instance 😄

sanjayprabhu: It's fine to panic if both are defined, that should not happen. This is mainly to make migration to the new config easier. If the logic is too complex, then I think we can migrate all existing configs, but that makes deployment more complicated. But I think it's easier in code? What do you think?

src/consensus/validator.rs
FrederikBolding: Done in https://github.com/farcasterxyz/snapchain/pull/279/commits/b7f04b5c674698afe47f34029a058e91996e578f

src/consensus/consensus.rs
FrederikBolding: Ended up just defaulting to the new format if both are defined and adding backwards compat in https://github.com/farcasterxyz/snapchain/pull/279/commits/b7f04b5c674698afe47f34029a058e91996e578f

sanjayprabhu (approved):
Looks good, thank you! Please rebase, and I'll merge.

sanjayprabhu:
I had to merge some things over the weekend that caused conflicts. I'll fix the conflicts and merge this.

PR #280, opened 2025-03-02
feat: Verify commit signatures when processing decided values

FrederikBolding:
Verify commit signatures when a read validator processes decided values. This verifies the signatures of the decided values and that the signers are expected for the height.

### DIFF
--- a/src/consensus/read_validator.rs
+++ b/src/consensus/read_validator.rs
@@ -135,12 +135,12 @@ impl ReadValidator {
     }
 
     pub fn process_decided_value(&mut self, value: DecidedValue) -> u64 {
+        let height = Self::get_decided_value_height(&value);
         let verified = Self::verify_signatures(&value);
         if !verified {
-            error!("Dropping decided block because its signatures are invalid");
+            error!(%height, last_height = %self.last_height, "Dropping decided block because its signatures are invalid");
             return 0;
         }
-        let height = Self::get_decided_value_height(&value);
         let num_committed_values = if height > self.last_height.increment() {
             if (self.buffered_blocks.len() as u32) < self.max_num_buffered_blocks {
                 self.buffered_blocks.insert(height, value);
--- a/src/consensus/read_validator.rs
+++ b/src/consensus/read_validator.rs
@@ -5,7 +5,7 @@ use crate::proto::{self, DecidedValue, Height};
 use crate::storage::store::engine::{BlockEngine, ShardEngine};
 use crate::utils::statsd_wrapper::StatsdClientWrapper;
 use bytes::Bytes;
-use informalsystems_malachitebft_core_types::NilOrVal;
+use informalsystems_malachitebft_core_types::{NilOrVal, ThresholdParams};
 use informalsystems_malachitebft_sync::RawDecidedValue;
 use itertools::Itertools;
 use libp2p::identity::ed25519::PublicKey;
@@ -138,14 +138,19 @@ impl ReadValidator {
             .map(|validator| validator.public_key.to_bytes());
 
         if certificate.aggregated_signature.signatures.len() == 0
-            || certificate.aggregated_signature.signatures.len() != expected_pubkeys.len()
+            || ThresholdParams::default().quorum.is_met(
+                certificate.aggregated_signature.signatures.len() as u64,
+                expected_pubkeys.len() as u64,
+            )
         {
+            error!(%certificate.height, last_height = %self.last_height, "Block did not have quorum");
             return false;
         }
 
         for signature in certificate.aggregated_signature.signatures {
             let address_bytes = &signature.address.0;
             if !expected_pubkeys.contains(address_bytes) {
+                error!(%certificate.height, last_height = %self.last_height, "Block contained signatures from unexpected signers");
                 return false;
             }
 
@@ -158,6 +163,7 @@ impl ReadValidator {
 
             let public_key = PublicKey::try_from_bytes(address_bytes).unwrap();
             if !public_key.verify(&vote.to_sign_bytes(), &signature.signature.0) {
+                error!(%certificate.height, last_height = %self.last_height, "Block contained invalid signatures");
                 return false;
             }
         }
--- a/src/consensus/read_validator.rs
+++ b/src/consensus/read_validator.rs
@@ -1,11 +1,13 @@
 use std::collections::BTreeMap;
 
-use crate::core::types::SnapchainValidatorContext;
+use crate::core::types::{SnapchainValidatorContext, Vote};
 use crate::proto::{self, DecidedValue, Height};
 use crate::storage::store::engine::{BlockEngine, ShardEngine};
 use crate::utils::statsd_wrapper::StatsdClientWrapper;
 use bytes::Bytes;
+use informalsystems_malachitebft_core_types::NilOrVal;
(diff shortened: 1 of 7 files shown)


src/consensus/read_validator.rs
```diff
@@ -96,7 +98,41 @@ impl ReadValidator {
...
+                .as_ref()
+                .unwrap()
+                .to_commit_certificate(),
+
+            proto::decided_value::Value::Block(block) => {
+                block.commits.as_ref().unwrap().to_commit_certificate()
+            }
+        };
+
+        for signature in certificate.aggregated_signature.signatures {
```
FrederikBolding, line 114: Wondering if 0 signatures probably should be treated as invalid too?

sanjayprabhu: Yeah, it should be invalid

src/consensus/read_validator.rs
```diff
@@ -96,7 +98,41 @@ impl ReadValidator {
...
+            let vote = Vote::new_precommit(
+                certificate.height,
+                certificate.round,
+                NilOrVal::Val(certificate.value_id.clone()),
+                signature.address.clone(),
+            );
+
+            let public_key = PublicKey::try_from_bytes(&signature.address.0).unwrap();
+            if !public_key.verify(&vote.to_sign_bytes(), &signature.signature.0) {
```
sanjayprabhu, line 123: We need to add one more check. This should use your validator set config from your other PR, and ensure that the public keys are correct for the height being verified.

src/consensus/read_validator.rs
```diff
@@ -96,7 +98,41 @@ impl ReadValidator {
...
+            }
+        }
+
+        true
+    }
+
     pub fn process_decided_value(&mut self, value: DecidedValue) -> u64 {
+        let verified = Self::verify_signatures(&value);
+        if !verified {
+            return 0;
```
sanjayprabhu, line 134: We should log an error here.

src/consensus/read_validator.rs
```diff
@@ -96,7 +98,41 @@ impl ReadValidator {
         }
     }
 
+    fn verify_signatures(value: &proto::DecidedValue) -> bool {
```
sanjayprabhu, line 101: Please add a test for this

FrederikBolding: Added signatures to all tests here https://github.com/farcasterxyz/snapchain/pull/280/commits/b91169a89dae9b948190985ba8ccd8a44bb7b97b

Let me know if you want something to further test this function, could do a unit test to make sure unsigned commits are skipped

src/consensus/read_validator_test.rs
```diff
@@ -143,4 +144,58 @@ mod tests {
...
+        });
+        let num_processed = process_decided_value(&mut read_validator, &shard_chunk).await;
+        assert_eq!(num_processed, 0);
+    }
+
+    #[tokio::test]
+    async fn test_ignore_invalid_signatures() {
+        let (mut proposer_engine, _read_node_engine, mut read_validator, proposer_keypair) =
+            setup(0).await;
+        let mut shard_chunk = commit_shard_chunk(&mut proposer_engine, &proposer_keypair).await;
```
sanjayprabhu, line 177: Might be a better test to sign with a brand new key?

src/consensus/read_validator.rs
```diff
@@ -96,7 +98,46 @@ impl ReadValidator {
...
+            }
+        }
+
+        true
+    }
+
     pub fn process_decided_value(&mut self, value: DecidedValue) -> u64 {
+        let verified = Self::verify_signatures(&value);
+        if !verified {
+            error!("Dropping decided block because its signatures are invalid");
```
sanjayprabhu, line 138: Let's add the height and the shard hash to the log.

sanjayprabhu (approved):

src/consensus/read_validator.rs
FrederikBolding: Done!

src/consensus/read_validator_test.rs
FrederikBolding: Done!

src/consensus/read_validator.rs
```diff
@@ -98,8 +104,74 @@ impl ReadValidator {
         }
     }
 
+    fn get_validator_set(&self, height: u64) -> SnapchainValidatorSet {
+        let mut result = &self.validator_set[0];
+        for config in &self.validator_set {
+            if config.shard_ids.contains(&self.shard_id)
+                && config.effective_at <= height
+                && config.effective_at > result.effective_at
```
sanjayprabhu, line 112: This logic is duplicated in the shard validator as well. I think it might be good to push this into the `StoredValidatorSet` (or create another struct that holds it) so it stores the vector inside, and we can just ask it for the validator set for a height

src/consensus/read_validator.rs
```diff
@@ -98,8 +104,74 @@ impl ReadValidator {
...
+
+        let validator_set = self.get_validator_set(certificate.height.as_u64());
+
+        let mut expected_pubkeys = validator_set
+            .validators
+            .iter()
+            .map(|validator| validator.public_key.to_bytes());
+
+        if certificate.aggregated_signature.signatures.len() == 0
+            || certificate.aggregated_signature.signatures.len() != expected_pubkeys.len()
```
sanjayprabhu, line 141: It won't always be equal.  With tendermint you need at least 2/3rds of the validators to sign. I think you should use Malachite's `ThresholdParam::default().quorom.is_met(signatures_length, validators_length) `

https://github.com/informalsystems/malachite/blob/53a9d9e071e773ff959465f2836648d8ad2a5c74/code/crates/core-types/src/threshold.rs#L34

src/consensus/read_validator.rs
```diff
@@ -98,8 +104,74 @@ impl ReadValidator {
...
+        {
+            return false;
+        }
+
+        for signature in certificate.aggregated_signature.signatures {
+            let address_bytes = &signature.address.0;
+            if !expected_pubkeys.contains(address_bytes) {
+                return false;
```
sanjayprabhu, line 149: For all the branches we return false, we should log an error explaining why. These branches should never be hit, so we should fail loudly.

src/consensus/read_validator.rs
FrederikBolding: My bad, fixed!

src/consensus/read_validator.rs
FrederikBolding: Added additional logging

src/consensus/read_validator.rs
FrederikBolding: Done!

sanjayprabhu (approved):
Nice, did you test this with an actual read node? 

You can uncomment this https://github.com/farcasterxyz/snapchain/blob/main/docker-compose.mainnet.yml#L5-L7

and run a node locally to make sure that the blocks are accepted

FrederikBolding:
> Nice, did you test this with an actual read node?
> 
> You can uncomment this https://github.com/farcasterxyz/snapchain/blob/main/docker-compose.mainnet.yml#L5-L7
> 
> and run a node locally to make sure that the blocks are accepted

I did! This correctly validates the signatures for the first couple of blocks on mainnet according to my testing!

PR #282, opened 2025-03-03
fixes for snapshot upload

aditiharini:
1. Use config files for AWS credentials. Using the default AWS envvars conflicts with our deploy process. 
2. Move the snapshot upload to a background thread so it doesn't block the rpc from returning. 
3. Use multipart upload. The regular `put_object` api call was failing often.

### DIFF
--- a/src/cfg.rs
+++ b/src/cfg.rs
@@ -61,7 +61,7 @@ impl Default for Config {
             statsd: StatsdConfig::default(),
             trie_branching_factor: 16,
             l1_rpc_url: "".to_string(),
-            fc_network: FarcasterNetwork::Testnet,
+            fc_network: FarcasterNetwork::Devnet,
             snapshot: storage::db::snapshot::Config::default(),
             read_node: false,
         }
--- a/src/network/admin_server.rs
+++ b/src/network/admin_server.rs
@@ -16,7 +16,7 @@ use std::{io, path, process};
 use thiserror::Error;
 use tokio::sync::mpsc;
 use tonic::{Request, Response, Status};
-use tracing::{info, warn};
+use tracing::{error, info, warn};
 
 pub struct DbManager {
     db_dir: String,
@@ -108,26 +108,22 @@ impl MyAdminService {
     }
 
     async fn backup_and_upload(
-        &self,
+        fc_network: FarcasterNetwork,
+        snapshot_config: storage::db::snapshot::Config,
         shard_id: u32,
         db: Arc<RocksDB>,
         now: i64,
     ) -> Result<(), Status> {
         // TODO(aditi): Eventually, we should upload a metadata file. For now, just clear all existing snapshots on s3 and only keep 1 snapshot per shard
-        clear_snapshots(self.fc_network, &self.snapshot_config, shard_id)
+        clear_snapshots(fc_network, &snapshot_config, shard_id)
             .await
             .map_err(|err| Status::from_error(Box::new(err)))?;
-        let backup_dir = self.snapshot_config.backup_dir.clone();
+        let backup_dir = snapshot_config.backup_dir.clone();
         let tar_gz_path = RocksDB::backup_db(db, &backup_dir, shard_id, now)
             .map_err(|err| Status::from_error(Box::new(err)))?;
-        storage::db::snapshot::upload_to_s3(
-            self.fc_network,
-            tar_gz_path,
-            &self.snapshot_config,
-            shard_id,
-        )
-        .await
-        .map_err(|err| Status::from_error(Box::new(err)))?;
+        storage::db::snapshot::upload_to_s3(fc_network, tar_gz_path, &snapshot_config, shard_id)
+            .await
+            .map_err(|err| Status::from_error(Box::new(err)))?;
         Ok(())
     }
 }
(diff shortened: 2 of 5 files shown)


src/cfg.rs
```diff
@@ -61,7 +61,7 @@ impl Default for Config {
             statsd: StatsdConfig::default(),
             trie_branching_factor: 16,
             l1_rpc_url: "".to_string(),
-            fc_network: FarcasterNetwork::Testnet,
```
sanjayprabhu, line 64: Make sure to update the config for the live nodes before this is merged.

src/network/admin_server.rs
```diff
@@ -231,33 +227,56 @@ impl AdminService for MyAdminService {
...
-
-        let on_error = |err| {
-            if let Err(err) = std::fs::remove_dir_all(self.snapshot_config.backup_dir.clone()) {
-                info!("Unable to remove snapshot directory: {}", err.to_string());
+        let fc_network = self.fc_network.clone();
+        let snapshot_config = self.snapshot_config.clone();
+        let shard_stores = self.shard_stores.clone();
+        let block_store = self.block_store.clone();
+        tokio::spawn(async move {
+            std::fs::exists(snapshot_config.backup_dir.clone()).unwrap();
```
sanjayprabhu, line 235: We lose the if check?

src/storage/db/snapshot.rs
```diff
@@ -264,24 +293,73 @@ pub async fn upload_to_s3(
...
         let key = format!("{}/{}", upload_dir, file_name);
 
-        info!(key, "Uploading chunk to s3");
-        let byte_stream = ByteStream::from_path(entry.path()).await?;
-        let upload_result = s3_client
-            .put_object()
-            .key(key.clone())
+        let mut file = tokio::fs::File::open(entry.path()).await?;
+        let mut buffer = Vec::new();
+        file.read_to_end(&mut buffer).await?;
```
sanjayprabhu, line 298: We shouldn't read everything into memory, we're chunking by 5mb anyway, we should read 5mb at a time into the buffer

src/storage/db/snapshot.rs
```diff
@@ -264,24 +293,73 @@ pub async fn upload_to_s3(
...
-                key,
-                snapshot_config.s3_bucket
-            );
+            .await?;
+
+        let upload_id = create_res.upload_id().unwrap();
+        let mut parts = Vec::new();
+
+        // 5 MB is the minimum size
+        for (i, chunk) in buffer.chunks(5 * 1024 * 1024).enumerate() {
```
sanjayprabhu, line 311: Let's make this 5mb a constant

src/storage/db/snapshot.rs
```diff
@@ -264,24 +293,73 @@ pub async fn upload_to_s3(
...
+                }
+            }
         }
-        upload_result?;
+
+        info!(key, "Finished uploading snapshot to s3");
+
+        // Complete Multipart Upload
+        s3_client
+            .complete_multipart_upload()
```
sanjayprabhu, line 351: If any of the parts fail, do we need to cancel the multipart upload? Does it have implications for future snapshots?

src/storage/db/snapshot.rs
```diff
@@ -301,9 +379,10 @@ pub async fn upload_to_s3(
         .content_type("application/json")
         .send()
         .await;
+
     if let Err(err) = &upload_result {
```
sanjayprabhu, line 383: Let's emit a metric on completion tagged by success/failure and shard id

src/storage/db/snapshot.rs
aditiharini: Yeah, we call `abort_multipart_upload` above and that should allow future uploads to go ahead.

src/cfg.rs
aditiharini: Yep, this is done

src/storage/db/snapshot.rs
aditiharini: Done

src/storage/db/snapshot.rs
aditiharini: done

src/network/admin_server.rs
aditiharini: Ah yeah, fixed

src/storage/db/snapshot.rs
```diff
@@ -264,24 +296,79 @@ pub async fn upload_to_s3(
...
+                .bucket(snapshot_config.s3_bucket.clone())
+                .key(key.clone())
+                .upload_id(upload_id)
+                .part_number(part_number)
+                .body(ByteStream::from(buf.to_vec()))
+                .send()
+                .await
+            {
+                Err(err) => {
+                    info!(key, "Aborting multipart snapshot upload");
```
sanjayprabhu, line 333: warn?

src/storage/db/snapshot.rs
```diff
@@ -264,24 +296,79 @@ pub async fn upload_to_s3(
...
+                            .e_tag(upload_res.e_tag().unwrap())
+                            .build(),
+                    );
+                }
+            }
         }
-        upload_result?;
+
+        info!(key, "Finished uploading snapshot to s3");
+        statsd_client.count_with_shard(shard_id, "snapshots.successful_upload", 1);
```
sanjayprabhu, line 356: Let's emit on failure as well.

sanjayprabhu (dismissed):

src/storage/db/snapshot.rs
aditiharini: There's a metric for failure above `snapshots.failed_upload`.

PR #291, opened 2025-03-05
feat: add missing methods

CassOnMars:


sanjayprabhu (approved):
Nice. Let's also add `submitMessage` to the http endpoint (here or in a separate PR).

PR #292, opened 2025-03-05
enable snapshot download

aditiharini:
Enable and snapshot download on startup. 
- Switched back from multipart upload to `put_object` because multipart upload was causing issues with decompression
- It's key to decompress the chunks into a single file before unpacking
- Added support for resuming onchain event/fname ingest from current place because past data should be included in blocks already. 

Did a lot of manual testing locally and on cluster 2:
- Started a validator node from snapshot
- Started a read node from snapshot

### DIFF
--- a/src/connectors/onchain_events/mod.rs
+++ b/src/connectors/onchain_events/mod.rs
@@ -162,6 +162,7 @@ pub struct Subscriber {
     stop_block_number: Option<u64>,
     statsd_client: StatsdClientWrapper,
     local_state_store: LocalStateStore,
+    start_from_current: bool,
 }
 
 // TODO(aditi): Wait for 1 confirmation before "committing" an onchain event.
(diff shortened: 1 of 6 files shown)


src/connectors/onchain_events/mod.rs
```diff
@@ -581,46 +618,28 @@ impl Subscriber {
...
-            latest_block_on_chain.min(self.stop_block_number.unwrap_or(latest_block_on_chain));
-
-        self.sync_historical_events(historical_sync_start_block, historical_sync_stop_block)
-            .await?;
+        match self.start_block_number {
+            None => {
+                self.sync_live_events(latest_block_on_chain).await?;
+            }
+            Some(start_block_number) => {
+                let historical_sync_start_block = latest_block_in_db.max(start_block_number);
```
sanjayprabhu, line 626: We should also do a `max(start_block_number, FIRST_BLOCK)` just to make sure we don't accidentally scan from 0 for blocks. We should never have to look at bloks older than FIRST_BLOCK

src/storage/db/rocksdb.rs
```diff
@@ -206,7 +202,7 @@ impl RocksDB {
 
         let mut multi_chunk_writer = MultiChunkWriter::new(
             PathBuf::from(chunked_output_dir.clone()),
-            4 * 1024 * 1024 * 1024, // 4GB
+            100 * 1024 * 1024, // 100MB
```
sanjayprabhu, line 205: Could you add a comment about why this number was picked?

sanjayprabhu (approved):

PR #301, opened 2025-03-12
enable snapshots for read nodes

aditiharini:
Change the config for the read node docker-compose to enable snapshot download.

### DIFF
--- a/docker-compose.read.yml
+++ b/docker-compose.read.yml
@@ -2,9 +2,9 @@ services:
   snap_read:
     image: farcasterxyz/snapchain:latest
     pull_policy: always
-#    build: # For testing
-#      context: .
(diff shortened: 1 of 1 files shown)


sanjayprabhu (approved):

sanjayprabhu:
Does it fallback to sync if snapshot fails or does it crash? We might want to have it fall back to sync

aditiharini:
> Does it fallback to sync if snapshot fails or does it crash? We might want to have it fall back to sync

It crashes. I think? this is ok for now-- I think it makes sense to look at the error and handle accordingly. If it was an early transient error, it probably makes sense to just restart once more and try to download the snapshot. I'll merge this as-is, but we can change if we want to.

PR #312, opened 2025-03-19
Revert "enable snapshots for read nodes (#301)"

aditiharini:
This reverts commit 97495feb5cbff2a8646f980267c0dcfab6f00362.

### DIFF
--- a/docker-compose.read.yml
+++ b/docker-compose.read.yml
@@ -2,9 +2,9 @@ services:
   snap_read:
     image: farcasterxyz/snapchain:latest
     pull_policy: always
-    #    build: # For testing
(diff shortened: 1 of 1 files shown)


docker-compose.read.yml
```diff
@@ -37,8 +37,8 @@ services:
         num_shards = 2
 
         [snapshot]
-        endpoint_url = "https://e1f9f185c6e63471dd39f96abd3413c4.r2.cloudflarestorage.com"
-        load_db_from_snapshot=true
+        endpoint_url = ""
```
sanjayprabhu, line 40: You can leave the endpoint url as is

PR #318, opened 2025-03-19
fix: Add retries for onchain events

sanjayprabhu:
Temporary connectivity issues should not halt onchain event processing

Tested manually

### DIFF
--- a/src/connectors/onchain_events/mod.rs
+++ b/src/connectors/onchain_events/mod.rs
@@ -58,6 +58,8 @@ const FIRST_BLOCK: u64 = 108864739;
 static CHAIN_ID: u32 = 10; // OP mainnet
 const RENT_EXPIRY_IN_SECONDS: u64 = 365 * 24 * 60 * 60; // One year
 
+const RETRY_TIMEOUT_SECONDS: u64 = 10;
+
 #[derive(Debug, Serialize, Deserialize)]
 pub struct Config {
     pub rpc_url: String,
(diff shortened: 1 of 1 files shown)


src/connectors/onchain_events/mod.rs
```diff
@@ -295,12 +295,36 @@ impl Subscriber {
...
+                Ok(Some(block)) => {
+                    return Ok(block.header.timestamp);
+                }
+                Ok(None) => {
+                    return Err(SubscribeError::UnableToFindBlockByHash);
+                }
+                Err(err) => {
+                    retry_count += 1;
+                    // Calculate exponential backoff with max of 60 seconds
+                    let backoff_secs = (2_u64.pow(retry_count.min(5) as u32)).min(60);
```
aditiharini, line 314: Why is there the .min(5)?

src/connectors/onchain_events/mod.rs
```diff
@@ -561,17 +626,36 @@ impl Subscriber {
     }
 
     async fn latest_block_on_chain(&mut self) -> Result<u64, SubscribeError> {
```
aditiharini, line 628: Maybe it's worth having just 1 fn with the retry logic and feeding the actual rpc call in via an anon fn or closure. Might be annoying to deal with this in rust though, it's ok to leave as is.

src/connectors/onchain_events/mod.rs
```diff
@@ -620,31 +704,52 @@ impl Subscriber {
...
+            return Ok(());
+        }
 
-                if should_start_live_sync {
-                    self.sync_live_events(historical_sync_stop_block).await?;
+        loop {
+            match self.sync_live_events(live_sync_block.unwrap()).await {
+                _ => {
+                    error!("Live sync ended unexpectedly. Retrying in 10 seconds",);
+                    tokio::time::sleep(tokio::time::Duration::from_secs(10)).await;
```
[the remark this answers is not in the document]
sanjayprabhu: I don't think we should panic, but I did make it stop retrying after a few attempts

src/connectors/onchain_events/mod.rs
```diff
@@ -620,31 +704,52 @@ impl Subscriber {
...
+        if live_sync_block.is_none() {
+            info!("Historical sync complete. Not subscribing to live events");
+            return Ok(());
+        }
 
-                if should_start_live_sync {
-                    self.sync_live_events(historical_sync_stop_block).await?;
+        loop {
+            match self.sync_live_events(live_sync_block.unwrap()).await {
+                _ => {
```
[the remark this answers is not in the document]
sanjayprabhu: Done

src/connectors/onchain_events/mod.rs
sanjayprabhu: I'd have to ask Claude :)

But I just removed this and used a simple constant time retry instead of exponential.

PR #319, opened 2025-03-20
fix: expedite processing of validator messages through mempool

aditiharini:
Prioritize processing onchain events and fname transfers so that we don't reject user messages due to missing onchain events.

### DIFF
--- a/src/mempool/mempool.rs
+++ b/src/mempool/mempool.rs
@@ -6,6 +6,7 @@ use std::{
 };
 use tokio::sync::{broadcast, mpsc, oneshot};
 
+use crate::core::util::farcaster_time_to_unix_seconds;
 use crate::{
     core::types::SnapchainValidatorContext,
     network::gossip::GossipEvent,
@@ -55,7 +56,7 @@ pub enum MempoolMessageKind {
 #[derive(Debug, PartialEq, Eq, PartialOrd, Ord)]
 pub struct MempoolKey {
     message_kind: MempoolMessageKind,
-    timestamp: u64,
+    timestamp: u64, // in unix seconds
     identity: String,
 }
 
@@ -79,7 +80,7 @@ impl proto::Message {
             // TODO: Consider revisiting choice of timestamp here as backdated messages currently are prioritized.
             return MempoolKey::new(
                 MempoolMessageKind::UserMessage,
-                data.timestamp as u64,
+                farcaster_time_to_unix_seconds(data.timestamp as u64),
                 self.hex_hash(),
             );
         }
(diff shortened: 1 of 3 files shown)


src/main.rs
```diff
@@ -224,7 +225,10 @@ async fn main() -> Result<(), Box<dyn Error>> {
     let keypair = app_config.consensus.keypair().clone();
 
     let (system_tx, mut system_rx) = mpsc::channel::<SystemMessage>(100);
-    let (mempool_tx, mempool_rx) = mpsc::channel(app_config.mempool.queue_size as usize);
+    let (user_message_mempool_tx, user_message_mempool_rx) =
+        mpsc::channel(app_config.mempool.queue_size as usize);
+    let (system_message_mempool_tx, system_message_mempool_rx) =
+        mpsc::channel(app_config.mempool.queue_size as usize);
```
sanjayprabhu, line 231: Why do we need two separate channels? So we don't drop messages if it's full?

aditiharini: Yeah, this is more relevant when the mempool size is limited. When that's the case, I didn't want an overload of user messages to result in onchain events being dropped or stalled.

aditiharini: When the mempool size is unlimited, the only thing that matters is the order we read out of the mempool in.

src/mempool/mempool.rs
```diff
@@ -69,7 +77,11 @@ impl proto::Message {
     pub fn mempool_key(&self) -> MempoolKey {
         if let Some(data) = &self.data {
             // TODO: Consider revisiting choice of timestamp here as backdated messages currently are prioritized.
-            return MempoolKey::new(data.timestamp as u64, self.hex_hash());
+            return MempoolKey::new(
+                MempoolMessageKind::UserMessage,
+                data.timestamp as u64,
```
sanjayprabhu, line 82: We should convert this to unix epoch so the timestamps are consistent for all messages in the mempool

src/mempool/mempool_test.rs
```diff
@@ -160,6 +160,82 @@ mod tests {
...
 
+    #[tokio::test]
+    async fn test_mempool_prioritization() {
+        let (_, _, mut mempool, mempool_tx, messages_request_tx, _shard_decision_tx, _) =
+            setup(setup_config(9304));
+
+        // Spawn mempool task
+        tokio::spawn(async move {
+            mempool.run().await;
+        });
```
sanjayprabhu, line 171: Does this keep running after the test ends? Do we need to abort the thread?

src/mempool/mempool.rs
aditiharini: done, converted to unix seconds which is consistent with fname server timestamp and block timestamp

sanjayprabhu (approved):

PR #321, opened 2025-03-20
feat: read node block pruning

dxj00:
wip for initial feedback on high level approach:

- Add `schedule_background_jobs` to main
- Add a config option `read_node_block_retention`
- Add `prune_until` to `src/storage/store/block.rs`


src/main.rs
```diff
@@ -129,6 +131,66 @@ async fn start_servers(
...
+}
+
+// TODO: handle shutdown_tx
+// TODO: does this need to be async?
+async fn schedule_background_jobs(app_config: &snapchain::cfg::Config, block_store: BlockStore) {
+    let sched = JobScheduler::new().await.unwrap();
+    if app_config.read_node {
+        if let Some(block_retention) = app_config.read_node_block_retention {
+            let cutoff_timestamp = (std::time::SystemTime::now() - block_retention)
+                .duration_since(std::time::UNIX_EPOCH)
```
sanjayprabhu, line 181: Header timestamps are in farcaster time (which uses a different epoch). You need to convert from farcaster time before you can compare.

src/storage/store/block.rs
```diff
@@ -217,4 +217,51 @@ impl BlockStore {
...
+        Ok(())
+    }
+
+    // Prune one page of blocks as specified by page_options from height 0 until
+    // but excluding the first block that matches the condition `f`. Returns the
+    // number of blocks pruned, and a boolean which is true if iteration stopped
+    // due to a block matching the condition, false otherwise.
+    pub fn prune_until<F>(
+        &self,
+        page_options: &PageOptions,
```
sanjayprabhu, line 233: I would move pagination into this function instead of the job.

Also, this is not on main yet, but will be merged later today, but we have an index on block by timestamp now https://github.com/farcasterxyz/snapchain/pull/315

So, you could could just provide a height to this function and make it simpler.

src/storage/store/block.rs
```diff
@@ -217,4 +217,51 @@ impl BlockStore {
...
+                .as_ref()
+                .ok_or(BlockStorageError::BlockMissingHeader)?;
+            if !f(&header) {
+                let height = header
+                    .height
+                    .as_ref()
+                    .ok_or(BlockStorageError::BlockMissingHeight)?;
+                let primary_key = make_block_key(height.block_number);
+                txn.delete(primary_key);
+                count += 1;
```
sanjayprabhu, line 255: We should add some throttling for the deletion.

src/storage/store/block.rs
dxj00: if we send pagination to this function then we should likely send a cancellation mechanism along with it.
do you have an example of a cancellation in this code base I can refer to?

src/lib.rs
```diff
@@ -14,3 +14,17 @@ mod tests;
 pub mod proto {
     tonic::include_proto!("_");
 }
+
+pub mod time {
+    use crate::core::types::FARCASTER_EPOCH;
+
+    pub fn current_time() -> u64 {
```
dxj00, line 21: not sure if this is the best place for this fn

src/background_jobs.rs
```diff
@@ -0,0 +1,59 @@
...
+    })
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use tokio::time::{sleep, Duration};
+
+    #[tokio::test]
+    async fn test_job_block_pruning() {
```
dxj00, line 43: this is just a test to see if I am using JobScheduler correctly, can ignore it for now

src/storage/store/block.rs
```diff
@@ -230,4 +235,215 @@ impl BlockStore {
             stop_block_number,
         )
     }
+
+    // Returns the next block height with a timestamp greater than or equal to
+    // the given timestamp for the specified shard index.
+    pub fn get_next_height_by_timestamp(
```
dxj00, line 241: this could be its own PR if desired

src/consensus/read_validator.rs
```diff
@@ -32,8 +32,10 @@ impl ReadValidator {
     }
 
     pub fn get_min_height(&self) -> Height {
-        // Always return the genesis block, until we implement pruning
-        Height::new(self.shard_id, 1)
+        match &self.engine {
+            Engine::BlockEngine(engine) => engine.get_min_height(),
+            Engine::ShardEngine(engine) => engine.get_min_height(),
+        }
```
dxj00, line 38: is `ReadValidator` the only place this needs to change? (does ReadValidator correspond to the setting read_node?)

sanjayprabhu: For now, yes. Eventually validators will also need similar pruning logic. 

And yes, ReadValidator is only used when read_node is set

src/lib.rs
sanjayprabhu: We should already have this in `core::util`?

src/storage/store/engine.rs
```diff
@@ -1331,6 +1331,11 @@ impl ShardEngine {
             })
     }
 
+    pub fn get_min_height(&self) -> Height {
+        // Always return the genesis block, until we implement pruning
+        Height::new(self.shard_id, 1)
```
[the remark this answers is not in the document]
sanjayprabhu: Yes, we need to prune the shard chunks as well. So we need similar logic in the shard store.

src/storage/store/block.rs
```diff
@@ -203,6 +224,21 @@ impl BlockStore {
         }
     }
 
+    #[inline]
+    pub fn min_block_number(&self) -> Result<u64, BlockStorageError> {
+        let first_block = self.get_first_block()?;
+        match first_block {
+            None => Ok(0),
```
aditiharini, line 231: The min block number should be 1 not 0

src/storage/store/block.rs
```diff
@@ -238,4 +274,206 @@ impl BlockStore {
...
+        shard_index: u32,
+        timestamp: u64,
+    ) -> Result<Option<u64>, BlockStorageError> {
+        let timestamp_index_key = make_block_timestamp_index(shard_index, timestamp);
+        let mut block_key: Option<Vec<u8>> = None;
+        self.db
+            .for_each_iterator_by_prefix_paged(
+                Some(timestamp_index_key),
+                None,
+                &PageOptions::default(),
```
aditiharini, line 291: You should set this to pull only 1 item from the db

src/storage/store/block.rs
```diff
@@ -238,4 +274,206 @@ impl BlockStore {
...
+            total_pruned += count;
+
+            if count == 0 {
+                break; // No more blocks to prune
+            }
+            info!("Pruning blocks... blocks pruned: {}", total_pruned);
+
+            // Sleep for the specified throttle duration
+            if throttle > Duration::zero() {
+                std::thread::sleep(throttle.to_std().unwrap());
```
aditiharini, line 372: Is there a reason for using a sync vs async function here? Not opposed, curious about reasoning.

aditiharini:
It's worth validating that `get_blocks` and `get_shard_chunks` still work properly with pruning.

aditiharini:
Read nodes still need to sync blocks from the beginning of history to make sure they build up the correct state in the trie and other stores. You should validate that they start syncing from the beginning and only prune after syncing. I would do this by just running a read node manually with the changes in this feature.

src/storage/store/block.rs
dxj00: happy to change it, I modeled this after `max_block_number` which also returns 0 in case there are no blocks in the db. should that be changed as well?

src/storage/store/block.rs
dxj00: not sure what the style of this code base is when using sync/async. I used sync since it seemed the other methods on `BlockStore` were sync. I agree this method is a bit different, lmk if you want me to change it to async.

PR #323, opened 2025-03-20
feat: Add auth for rpc endpoints and cleanup admin rpc

sanjayprabhu:
Add auth for submit and subscribe rpc endpoints. And for admin endpoints. Don't mount the admin endpoint if admin rpc auth is not configured. Also remove admin endpoints that directly mutate state, we shouldn't allow this for mainnet


src/proto/admin_rpc.proto
```diff
@@ -5,16 +5,8 @@ import "username_proof.proto";
...
-
-message TerminateResponse {
-}
-
 service AdminService {
-  rpc Terminate(TerminateRequest) returns (TerminateResponse);
-  rpc SubmitOnChainEvent(OnChainEvent) returns (OnChainEvent);
-  rpc SubmitUserNameProof(UserNameProof) returns (UserNameProof);
+//  rpc SubmitOnChainEvent(OnChainEvent) returns (OnChainEvent);
+//  rpc SubmitUserNameProof(UserNameProof) returns (UserNameProof);
```
sanjayprabhu, line 10: These were meant for debugging and should not be allowed on mainnet. Remove for now, we can add back in a `DebugService` later if required (which is not mounted if the configured network is mainnet).

PR #327, opened 2025-03-21
fix: fix mempool logging for link compact messages

aditiharini:
We log an error consistently for link compact messages in the mempool despite processing them. Remove the log.

### DIFF
--- a/src/mempool/mempool.rs
+++ b/src/mempool/mempool.rs
@@ -177,8 +177,8 @@ impl ReadNodeMempool {
                     Some(message_data) => {
                         let ts_hash = make_ts_hash(message_data.timestamp, &message.hash).unwrap();
                         match type_to_set_postfix(message_data.r#type()) {
-                            Err(err) => {
-                                error!("Error retrieving set postfix: {}", err.to_string());
+                            Err(_) => {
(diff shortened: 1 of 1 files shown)


src/mempool/mempool.rs
```diff
@@ -177,8 +177,8 @@ impl ReadNodeMempool {
                     Some(message_data) => {
                         let ts_hash = make_ts_hash(message_data.timestamp, &message.hash).unwrap();
                         match type_to_set_postfix(message_data.r#type()) {
-                            Err(err) => {
-                                error!("Error retrieving set postfix: {}", err.to_string());
+                            Err(_) => {
+                                // We hit this for link compact messages and it's expected. Return [false] so the message isn't filtered out
```
sanjayprabhu, line 181: Why is it expected?

sanjayprabhu (approved):

src/mempool/mempool.rs
aditiharini: Good question, the `type_to_set_postfix` already existed and didn't account for link compact messages. It's used elsewhere so i didn't want to change semantics. It might just be more correct for that function to handle link compact messages. I think it's safer to merge this as-is right now and fix that later if we want to.

PR #330, opened 2025-03-21
feat: read node block pruning

dxj00:
wip for initial feedback on high level approach:

- Add `schedule_background_jobs` to main
- Add a config option `read_node_block_retention`
- Add `prune_until` to `src/storage/store/block.rs`

Migrated from: https://github.com/farcasterxyz/snapchain/pull/321 (has some comments/discussion)


src/storage/store/shard.rs
```diff
@@ -255,4 +289,60 @@ impl ShardStore {
...
+        stop_height: u64,
+        page_options: &PageOptions,
+        throttle: Duration,
+        shutdown_rx: Option<watch::Receiver<()>>,
+    ) -> Result<u32, ShardStorageError> {
+        let stop_prefix = Some(make_shard_key(stop_height));
+        let total_pruned = self
+            .db
+            .delete_paginated(
+                None,
```
dxj00, line 334: does this need to be `Some(make_shard_key(0))`?

src/storage/store/shard.rs
```diff
@@ -208,6 +228,20 @@ impl ShardStore {
         }
     }
 
+    pub fn min_block_number(&self) -> Result<u64, ShardStorageError> {
+        let first_shard_chunk = get_first_or_last_shard_chunk(&self.db, FirstOrLast::First)?;
+        match first_shard_chunk {
+            None => Ok(0),
```
dxj00, line 234: still unsure about returning 0/1 here. maybe this should be handled at the callsite?

dxj00: for now I put a band-aid on the callsite as max(1)

src/storage/store/shard.rs
dxj00: modified it

src/main.rs
```diff
@@ -274,6 +307,18 @@ async fn main() -> Result<(), Box<dyn Error>> {
             Err(_) => None,
         };
 
+    let global_db = RocksDB::open_global_db(&app_config.rocksdb_dir);
+    let local_state_store = LocalStateStore::new(global_db);
+    let cancel = CancellationToken::new();
```
dxj00, line 312: I'm not sure use of a cancellation token is necessary as I don't think async jobs will continue past returning from main.

src/background_jobs.rs
```diff
@@ -0,0 +1,33 @@
...
+
+pub fn job_block_pruning(
+    schedule: &str,
+    cutoff_timestamp: u64,
+    throttle: Duration,
+    block_store: BlockStore,
+) -> Result<Job, JobSchedulerError> {
+    Job::new_async(schedule, move |_uuid, _l| {
+        let block_store = block_store.clone(); // Get Arc for this job (can this clone be avoided?)
+        Box::pin(async move {
```
sanjayprabhu, line 16: Why do we need the `Box::pin`?

src/main.rs
```diff
@@ -130,6 +132,34 @@ async fn start_servers(
     });
 }
 
+async fn schedule_background_jobs(
+    app_config: &snapchain::cfg::Config,
+    block_store: BlockStore,
+    local_state_store: LocalStateStore,
+) {
+    let sched = JobScheduler::new().await.unwrap();
+    let is_sync_complete = local_state_store.is_sync_complete().unwrap();
```
sanjayprabhu, line 141: I think this should go inside the job. Let's keep the scheduling logic as simple as possible.

src/main.rs
```diff
@@ -130,6 +132,34 @@ async fn start_servers(
...
+    if app_config.read_node && is_sync_complete {
+        if let Some(block_retention) = app_config.read_node_block_retention {
+            let throttle = tokio::time::Duration::from_millis(100); // TODO: make const or configurable
```
sanjayprabhu, line 144: Should also go inside the job

src/main.rs
```diff
@@ -130,6 +132,34 @@ async fn start_servers(
...
+            let cutoff_timestamp =
+                util::get_farcaster_time().unwrap() - (block_retention.as_secs() as u64);
```
sanjayprabhu, line 146: Let's pass in just the block_retention, and have the job calculate the timestamp

src/storage/store/shard.rs
sanjayprabhu: Hmm, I think it's more correct to return and Error if there are no blocks instead of having to do max(1). And and have the caller decide if it should propagate the error or return 0.

src/storage/db/rocksdb.rs
```diff
@@ -553,6 +554,79 @@ impl RocksDB {
 
         Ok(count)
     }
+
+    pub fn get_next_by_index(&self, start_prefix: Vec<u8>) -> Result<Option<Vec<u8>>, HubError> {
```
sanjayprabhu, line 558: We should add a test for this

src/main.rs
```diff
@@ -333,6 +368,7 @@ async fn main() -> Result<(), Box<dyn Error>> {
                             // [num_shards] doesn't account for the block shard, so account for it manually
                             if shards_finished_syncing.len() as u32 == app_config.consensus.num_shards + 1 {
                                 info!("Initial sync completed for all shards");
+                                local_state_store.set_sync_complete(true)?;
```
sanjayprabhu, line 371: I think this has to take in a shard id as well, and then keep track of this field per shard.

src/main.rs
sanjayprabhu: Yeah, we don't need it

src/main.rs
dxj00: I'm curious if this is necessary if this path is going to be triggered each time on init.
We could just schedule the job at this point instead of persisting a flag to the db.

sanjayprabhu: Yeah, we can have it trigger here. That works as well, and we don't need to store this. We should use a channel to send a message to the scheduler.

src/storage/store/shard.rs
```diff
@@ -255,4 +288,57 @@ impl ShardStore {
...
+    pub async fn prune_until(
+        &self,
+        stop_height: u64,
+        page_options: &PageOptions,
+        throttle: Duration,
+    ) -> Result<u32, ShardStorageError> {
+        let total_pruned = self
+            .db
+            .delete_paginated(
+                Some(make_shard_key(0)),
```
dxj00, line 331: I think I'm a little bit confused about why `make_shard_key` doesn't take the shard id? perhaps I'm not understanding the data store schema (key -> value mapping) correctly.

sanjayprabhu: Ah, yeah, I think that's a bug. It's not affecting us because each shard has a different db. But it shouldn't had a shard id in the key.

src/background_jobs.rs
```diff
@@ -0,0 +1,78 @@
...
+        let sync_complete = *sync_complete_rx.borrow();
+        if !sync_complete {
+            info!("Sync not complete, skipping block pruning");
+            return;
+        }
+
+        let cutoff_timestamp =
+            util::get_farcaster_time().unwrap() - (block_retention.as_secs() as u64);
+        let stop_height = block_store
+            .get_next_height_by_timestamp(0, cutoff_timestamp)
```
sanjayprabhu, line 30: We shouldn't require the shard id as a param here, since each store should know the shard if to use. What does it mean if the caller provides a shard id that doesn't match the store?

The shard stores don't hold the shard id, but let's provide it in the constructor so they can use it automatically.

src/background_jobs.rs
```diff
@@ -0,0 +1,78 @@
...
+                .unwrap_or_else(|e| {
+                    error!("Error pruning blocks: {}", e);
+                    0
+                });
+        });
+
+        for (shard_id, stores) in shard_stores.iter() {
+            let shard_store = &stores.shard_store;
+            let stop_height = shard_store
+                .get_next_height_by_timestamp(0, cutoff_timestamp)
```
sanjayprabhu, line 55: 0 is the wrong shard id here, I think?

src/storage/db/rocksdb.rs
```diff
@@ -553,6 +554,79 @@ impl RocksDB {
...
+        let page_options = PageOptions {
+            page_size: Some(1),
+            ..PageOptions::default()
+        };
+        let mut primary_key: Option<Vec<u8>> = None;
+        self.for_each_iterator_by_prefix(Some(start_prefix), None, &page_options, |_, index| {
```
sanjayprabhu, line 564: Hmm, I think there's a subtle bug here. If the key doesn't exist, it will overflow onto the next prefix. 

I think we need to pass in only `[RootPrefix::BlockIndex, shard_index]` and then call increment_vec_u8 as the stop prefix, like here https://github.com/farcasterxyz/snapchain/blob/20335edb2a5557c92aba9331536cb54e313451e9/src/storage/store/account/link_store.rs#L139-L141 

This should prevent it from overflowing. Let's add a test for this as well.

sanjayprabhu (commented):
Looking good.

src/background_jobs.rs
dxj00: now it uses the shard_id from the struct 👍

src/storage/db/rocksdb.rs
dxj00: yep, good catch.

src/background_jobs.rs
dxj00: had to switch this back to be able to call `.prune_until(...).await`
as an alternative we could create a runtime like:
```
    let rt = Runtime::new().unwrap();
    let result = rt.block_on(.prune_until(...).await);
```
but this seems clearer (to use the async fn provided).

not sure if the clones can be avoided.

src/background_jobs.rs
```diff
@@ -0,0 +1,87 @@
...
+
+pub fn job_block_pruning(
+    schedule: &str,
+    block_retention: Duration,
+    block_store: BlockStore,
+    shard_stores: HashMap<u32, Stores>,
+    sync_complete_rx: watch::Receiver<bool>,
+) -> Result<Job, JobSchedulerError> {
+    Job::new_async(schedule, move |_, _| {
+        // TODO: Can these clones be avoided?
```
sanjayprabhu, line 20: Is the borrow checker complaining if you don't clone? They should be moved into the closure

sanjayprabhu (approved):
Would be nice to have a job level test that that would've caught the issues you found with manual testing. But not a big deal if you don't have time. 

Looks good to me.

src/background_jobs.rs
dxj00: yes, it generates this error:

```
cannot move out of `sync_complete_rx`, a captured variable in an `FnMut` closure
`sync_complete_rx` is moved hererustc[Click for full compiler diagnostic](rust-analyzer-diagnostics-view:/diagnostic%20message%20[0]?0#file:///Users/.../src/snapchain/src/background_jobs.rs)
background_jobs.rs(17, 5): captured outer variable
background_jobs.rs(19, 30): captured by this `FnMut` closure
background_jobs.rs(25, 34): variable moved due to use in coroutine
background_jobs.rs(25, 34): move occurs because `sync_complete_rx` has type `tokio::sync::watch::Receiver<bool>`, which does not implement the `Copy` trait
background_jobs.rs(23, 18): clone the value before moving it into the closure: `{
```
which I am not sure how to resolve.

PR #335, opened 2025-03-26
fix: Auto reconnect to bootstrap peers if connection is lost

sanjayprabhu:
Reconnect to bootstrap peers if connection is lost

### DIFF
--- a/src/network/gossip.rs
+++ b/src/network/gossip.rs
@@ -109,6 +109,7 @@ pub struct SnapchainGossip {
     system_tx: Sender<SystemMessage>,
     sync_channels: HashMap<InboundRequestId, sync::ResponseChannel>,
     read_node: bool,
+    bootstrap_addrs: Vec<String>,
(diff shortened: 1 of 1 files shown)


src/network/gossip.rs
```diff
@@ -175,20 +176,7 @@ impl SnapchainGossip {
...
-            info!("Dialing bootstrap peer: {:?} ({:?})", parsed_addr, addr);
-            let res = swarm.dial(opts);
-            if let Err(e) = res {
-                warn!(
-                    "Failed to dial bootstrap peer {:?}: {:?}",
-                    parsed_addr.clone(),
-                    e
-                );
-            }
+            let _ = Self::dial(&mut swarm, &addr);
```
aditiharini, line 179: Should log error if there is one.

sanjayprabhu: It's logged inside `dial`

PR #337, opened 2025-03-28
feat: support peer autodiscovery 

aditiharini:
Allow nodes to find each other so that they can receive data from nodes other than the bootstrap nodes.

### DIFF
--- a/src/network/gossip.rs
+++ b/src/network/gossip.rs
@@ -341,8 +341,10 @@ impl SnapchainGossip {
                     self.statsd_client.gauge("gossip.connected_peers", self.swarm.connected_peers().count() as u64);
                 },
                 _ = publish_contact_info_timer.tick() => {
-                    info!("Publishing contact info");
-                    self.publish_contact_info()
+                    if self.read_node {
+                        info!("Publishing contact info");
+                        self.publish_contact_info()
+                    }
                 }
                 gossip_event = self.swarm.select_next_some() => {
                     match gossip_event {
--- a/src/network/gossip_test.rs
+++ b/src/network/gossip_test.rs
@@ -48,7 +48,7 @@ async fn test_gossip_communication() {
         keypair1.clone(),
         &config1,
         system_tx1,
-        false,
+        true,
         FarcasterNetwork::Devnet,
         statsd_client(),
     )


src/network/gossip_test.rs
```diff
@@ -94,3 +130,72 @@ async fn test_gossip_communication() {
...
+        gossip1.start().await;
+    });
+    tokio::spawn(async move {
+        gossip2.start().await;
+    });
+    tokio::spawn(async move {
+        gossip3.start().await;
+    });
+
+    tokio::time::sleep(Duration::from_secs(1)).await;
```
sanjayprabhu, line 200: No asserts?

src/network/gossip.rs
```diff
@@ -391,12 +462,45 @@ impl SnapchainGossip {
...
+            return;
+        }
+
+        if contact_info_body.network() != self.fc_network {
+            return;
+        }
+
+        if contact_info_body.snapchain_version != FARCASTER_VERSION {
+            return;
+        }
```
sanjayprabhu, line 483: We should also return early for validators?

src/network/gossip_test.rs
aditiharini: Ah good catch, this was a draft function that I wrote. I ended up incorporating the behavior I wanted to test into the existing test.

src/network/gossip.rs
```diff
@@ -391,12 +462,45 @@ impl SnapchainGossip {
...
+
+        let _ = Self::dial(&mut self.swarm, &contact_info_body.gossip_address);
```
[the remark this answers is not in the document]
aditiharini: Added the `ConnectionLimits` to address

src/network/gossip.rs
aditiharini: I changed the code so that we only call this function for read nodes. There's no need for validators to autodiscover peers.

src/network/gossip.rs
```diff
@@ -248,18 +303,46 @@ impl SnapchainGossip {
...
                 },
+                _ = publish_contact_info_timer.tick() => {
+                    info!("Publishing contact info");
+                    self.publish_contact_info()
+                }
                 gossip_event = self.swarm.select_next_some() => {
                     match gossip_event {
                         SwarmEvent::ConnectionEstablished {peer_id, ..} => {
-                            info!("Connection established with peer: {peer_id}");
+                            info!(total_peers = self.swarm.connected_peers().count(), "Connection established with peer: {peer_id}");
```
sanjayprabhu, line 345: Let's emit a stat for this too, maybe in the reconnect timer.

src/network/gossip.rs
```diff
@@ -248,18 +303,46 @@ impl SnapchainGossip {
...
+                    .as_millis() as u64,
+            }),
+        };
+
+        let gossip_message = GossipMessage {
+            gossip_message: Some(gossip_message::GossipMessage::ContactInfoMessage(
+                contact_info,
+            )),
+        };
+        self.publish(gossip_message.encode_to_vec(), CONTACT_INFO);
```
sanjayprabhu, line 325: I think validators shouldn't publish contactInfo, we don't want to induce extra connection than what happens at bootstrap

sanjayprabhu (approved):

PR #338, opened 2025-03-28
fix: Return store errors on simulate and avoid logging

sanjayprabhu:
Reduces excessive logging on invalid messages and returns the store error back to the caller so they can handle appropriately.


src/storage/store/engine.rs
```diff
@@ -641,17 +645,21 @@ impl ShardEngine {
                                                 events.push(hub_event.clone());
                                             }
                                             Err(err) => {
-                                                warn!(
-                                                    "Error revoking existing username: {:?}",
-                                                    err
-                                                );
+                                                if source != ProposalSource::Simulate {
```
sanjayprabhu, line 648: Would've been cleaner to change the log level dynamically. But it requires some rust macro magic. Just went with an if check for now. We can cleanup later.

PR #341, opened 2025-03-31
fix: stop tracking proposals that will never be used again

aditiharini:
We're leaking memory in `proposed_blocks` and `proposed_shard_chunks` by not removing proposed values for older heights and rounds.

### DIFF
--- a/src/consensus/proposer.rs
+++ b/src/consensus/proposer.rs
@@ -90,18 +90,10 @@ impl ProposedValues {
     }
 
     pub fn decide(&mut self, height: Height) {
-        while let Some((entry_height, entry_shard_hashes)) = self.values_by_height.pop_first() {
-            if entry_height <= height {
-                for entry_shard_hash in entry_shard_hashes {
-                    self.values.remove(&entry_shard_hash);
-                }
-                continue;
+        if let Some(shard_hashes) = self.values_by_height.remove(&height) {
+            for shard_hash in shard_hashes {
+                self.values.remove(&shard_hash);
             }
-
-            // Put it back in, we shouldn't have removed
-            self.values_by_height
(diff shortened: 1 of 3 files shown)


src/consensus/proposer.rs
```diff
@@ -58,10 +58,56 @@ pub trait Proposer {
     fn get_proposed_value(&self, shard_hash: &ShardHash) -> Option<FullProposal>;
 }
 
+pub struct ProposedValues {
+    values_by_height: BTreeMap<(Height, Round), ShardHash>,
+    values: BTreeMap<ShardHash, FullProposal>,
+}
+
+impl ProposedValues {
```
sanjayprabhu, line 66: Worth adding a test for this. 

We should also add tests for the proposers themselves, but that might require a bunch of setup work. So, ok to defer for now

src/consensus/proposer.rs
```diff
@@ -73,28 +73,34 @@ impl ProposedValues {
...
+                for entry_shard_hash in entry_shard_hashes {
+                    self.values.remove(&entry_shard_hash);
+                }
                 continue;
             }
 
             // Put it back in, we shouldn't have removed
             self.values_by_height
-                .insert((entry_height, entry_round), entry_shard_hash);
+                .insert(entry_height, entry_shard_hashes);
```
sanjayprabhu, line 103: Hmm, this insert feels dangerous. There's a potential for a race condition here where we've added a value for a height and then this insert replaces it with an outdated vector.

I think we should iterate and collect the heights we want to remove first and then remove it in one go or always remove only the previous height if it exists. The latter seems cleaner?

sanjayprabhu: But I think it's more correct to remove all previous heights. So, would be better to do the former

PR #347, opened 2025-04-01
feat: Add a background job to prune events

sanjayprabhu:
Background job to delete events after 3 days (same as hubs)

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -148,10 +148,11 @@ async fn schedule_background_jobs(
     sync_complete_rx: watch::Receiver<bool>,
 ) {
     let sched = JobScheduler::new().await.unwrap();
+    let mut jobs = vec![];
     if app_config.read_node {
-        if let Some(block_retention) = app_config.read_node_block_retention {
+        if let Some(block_retention) = app_config.pruning.block_retention {
             let schedule = "0 0 0 * * *"; // midnight UTC every day
-            let job = snapchain::background_jobs::job_block_pruning(
+            let job = snapchain::jobs::block_pruning::block_pruning_job(
                 schedule,
                 block_retention,
                 block_store.clone(),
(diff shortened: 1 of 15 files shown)


src/storage/store/stores.rs
```diff
@@ -338,6 +347,63 @@ impl Stores {
...
+            );
+            let start = std::time::Instant::now();
+            let count =
+                HubEvent::prune_events_util(self.db.clone(), stop_height, &page_options, throttle)
+                    .await?;
+            let elapsed = start.elapsed();
+            self.statsd
+                .count_with_shard(self.shard_id, "prune.events", count as u64);
+            self.statsd.time_with_shard(
+                self.shard_id,
```
aditiharini, line 395: Can we add ms to the metric name? I think it'd be good to do this for all the timing metrics so it's easy to figure out the scale of the y axis.

src/main.rs
```diff
@@ -148,21 +148,33 @@ async fn schedule_background_jobs(
...
                 sync_complete_rx,
             )
             .unwrap();
-            sched.add(job).await.unwrap();
+            jobs.push(job);
         }
     }
-    // Other background jobs can be added here
+
+    let event_pruning_job = snapchain::jobs::event_pruning::event_pruning_job(
```
aditiharini, line 167: Do we want this job running at the exact same time on all validators? I think it'd make sense to stagger because this is an expensive operation.

aditiharini (dismissed):

src/storage/store/stores.rs
sanjayprabhu: Fixed

src/main.rs
sanjayprabhu: Based on testnet, this seems to be fine. The first prune will be taxing, but will make sure that happens after backfill is complete. If it becomes a problem, we can move to pruning on commit

PR #348, opened 2025-04-02
feat: rate limits in mempool

aditiharini:
Add rate limits in the mempool.

### DIFF
--- a/src/mempool/mempool.rs
+++ b/src/mempool/mempool.rs
@@ -54,7 +54,7 @@ impl Default for RateLimitsConfig {
         Self {
             time_to_idle: Duration::from_secs(60 * 2 * 2),
             // Make time to idle 2x the rate limit window so it's ok if out of sync
-            max_capacity: 10_000,
+            max_capacity: 1_000_000,
         }
     }
 }
@@ -94,6 +94,7 @@ impl RateLimits {
             if storage_allowance == 0 {
                 Err(Box::new(RateLimitsError::NoStorage))
             } else {
+                // If we update the quota, we should update [time_to_idle] accordingly
                 Ok(Arc::new(RateLimiter::direct(Quota::per_hour(
                     NonZeroU32::new(100.max(storage_allowance / 10)).unwrap(),
                 ))))
--- a/src/mempool/mempool.rs
+++ b/src/mempool/mempool.rs
@@ -1,3 +1,5 @@
+use governor::clock::QuantaClock;
+use governor::state::{InMemoryState, NotKeyed};
 use serde::{Deserialize, Serialize};
 use std::cmp::PartialEq;
 use std::{
@@ -7,6 +9,7 @@ use std::{
 use tokio::sync::{broadcast, mpsc, oneshot};
 
 use crate::core::util::farcaster_time_to_unix_seconds;
+use crate::proto::OnChainEventType;
 use crate::{
     core::types::SnapchainValidatorContext,
     network::gossip::GossipEvent,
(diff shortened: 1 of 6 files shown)


src/mempool/mempool.rs
```diff
@@ -26,14 +30,93 @@ use crate::{
...
+            let storage_limits = stores.get_storage_limits(fid).unwrap();
+            let storage_allowance: u32 = storage_limits
+                .limits
+                .iter()
+                .map(|limits| limits.limit as u32)
+                .sum();
+            if storage_allowance == 0 {
+                Err(Box::new(RateLimitsError::NoStorage))
+            } else {
+                Ok(Arc::new(RateLimiter::direct(Quota::per_hour(
```
sanjayprabhu, line 97: Add a comment here about updating the time_to_idle if we change the quota duration

src/mempool/mempool.rs
```diff
@@ -26,14 +30,93 @@ use crate::{
...
+    pub time_to_idle: Duration,
+    pub max_capacity: usize,
+}
+
+impl Default for RateLimitsConfig {
+    fn default() -> Self {
+        Self {
+            time_to_idle: Duration::from_secs(60 * 2 * 2),
+            // Make time to idle 2x the rate limit window so it's ok if out of sync
+            max_capacity: 10_000,
```
sanjayprabhu, line 57: This may be too low, this should be at least 100K, but I think we should be even more lenient and set it to 1M. Worst case is 1K msgs/block from unique fids * 1 hour in seconds = 3.6M fids. 

We should emit a stat on how much of this capcaity is used

sanjayprabhu: Hmm, this is global, and not shared per shard? Even more important to bump this up.

src/mempool/mempool_test.rs
```diff
@@ -172,9 +175,39 @@ mod tests {
         assert!(!valid)
     }
 
+    #[tokio::test]
+    async fn test_rate_limits_applied() {
+        let (mut engine, _, mut mempool, _, _, _, _) = setup(None, true);
```
sanjayprabhu, line 180: Hmm, the only difference is that we pass true here? Can we just pass trie everywhere and remove this test? What do we gain from setting it to false in other tests?

src/mempool/rate_limits_test.rs
```diff
@@ -0,0 +1,229 @@
...
+            stores::{Limits, StoreLimits, Stores},
+            test_helper::{self, register_user, EngineOptions, FID_FOR_TEST},
+        },
+    };
+
+    fn setup(limits: Limits) -> (ShardEngine, HashMap<u32, Stores>) {
+        let (engine, _tmpdir) = test_helper::new_engine_with_options(EngineOptions {
+            limits: Some(StoreLimits {
+                limits,
+                legacy_limits: Limits {
```
sanjayprabhu, line 20: There's a `limits::zero` in `test_helper`

src/mempool/rate_limits_test.rs
```diff
@@ -0,0 +1,229 @@
...
+        assert!(!rate_limits.consume_for_fid(engine.shard_id(), FID_FOR_TEST));
+
+        sleep(Duration::from_millis(1500)).await;
+
+        assert!(rate_limits.consume_for_fid(engine.shard_id(), FID_FOR_TEST));
+    }
+
+    #[tokio::test]
+    async fn test_exceed_time_to_idle() {
+        // Make limits high so refresh rate is high enough that we can send a message after the rate limit is hit in test.
```
sanjayprabhu, line 78: Is the comment correct?

src/mempool/mempool.rs
```diff
@@ -434,7 +508,17 @@ impl Mempool {
                                     }
                                     for system_message in transaction.system_messages {
                                         mempool.remove(&system_message.mempool_key());
-                                        self.statsd_client.count_with_shard(height.shard_index, "mempool.remove.success", 1);
+                                        if let Some(onchain_event) = system_message.on_chain_event
+                                        {
+                                            if onchain_event.r#type() == OnChainEventType::EventTypeStorageRent{
+                                                // If the user buys more storage, we should bump their rate limit
+                                                if let Some(rate_limits) = &mut self.rate_limits {
```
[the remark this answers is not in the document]
aditiharini: Discussed offline and agreed it's reasonable to keep.

sanjayprabhu (approved):

src/mempool/mempool_test.rs
aditiharini: It's a bit annoying to have to think about the rate limits in all the tests. They all fail right now if I enable the rate limits.

PR #350, opened 2025-04-04
fix: allow larger message size for link compact messages

aditiharini:
Link compact messages can exceed the allowed 2048 bytes. The max size of a link compact state message I could find was ~61kB so set the max to something a bit higher than that.

### DIFF
--- a/src/core/validations/message.rs
+++ b/src/core/validations/message.rs
@@ -9,6 +9,7 @@ use prost::Message;
 use super::{cast, link, reaction, verification};
 
 const MAX_DATA_BYTES: usize = 2048;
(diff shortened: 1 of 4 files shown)


src/core/validations/message.rs
```diff
@@ -9,6 +9,7 @@ use prost::Message;
 use super::{cast, link, reaction, verification};
 
 const MAX_DATA_BYTES: usize = 2048;
+const MAX_DATA_BYTES_FOR_LINK_COMPACT: usize = 65536;
```
sanjayprabhu, line 12: This is based on the db size?

sanjayprabhu (approved):

PR #351, opened 2025-04-04
fix: Ensure pruning jobs don't run into each other

sanjayprabhu:
Event and shard pruning jobs can run into each the first time they are run. Add a lock to prevent this.


src/storage/store/stores.rs
```diff
@@ -199,6 +201,7 @@ impl Stores {
             store_limits,
             event_handler,
             statsd,
+            prune_lock: Arc::new(RwLock::new(false)),
```
aditiharini, line 204: Is there a reason we can't use a mutex here?

aditiharini (dismissed):

PR #355, opened 2025-04-07
feat: add snapshot upload background job

aditiharini:
Add an automated snapshot upload job. 

Note: I needed to update dependency versions to fix the build.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -85,6 +85,7 @@ strum = "0.27.1"
 strum_macros = "0.27.1"
 governor = "0.10.0"
 moka = "0.12.10"
+// Foundry sets the version of solar-parse to 0.1.1 and these need to be consistent for compilation to succeed. Foundry doesn't have a cargo crate that we can use. 
 solar-interface = "=0.1.1"
 solar-ast = "=0.1.1"
 solar-data-structures = "=0.1.1"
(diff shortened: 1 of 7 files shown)


Cargo.toml
```diff
@@ -85,6 +85,11 @@ strum = "0.27.1"
 strum_macros = "0.27.1"
 governor = "0.10.0"
 moka = "0.12.10"
+solar-interface = "=0.1.1"
+solar-ast = "=0.1.1"
+solar-data-structures = "=0.1.1"
+solar-macros = "=0.1.1"
+solar-config = "=0.1.1"
```
sanjayprabhu, line 92: Cana we add a comment about why we need this?

src/main.rs
```diff
@@ -171,6 +172,17 @@ async fn schedule_background_jobs(
     .unwrap();
     jobs.push(event_pruning_job);
 
+    let snapshot_upload_job = snapchain::jobs::snapshot_upload::snapshot_upload_job(
+        "0 0 5 * * *", // 5 AM UTC every day
```
sanjayprabhu, line 176: We should only do this if snapshots are enabled.

aditiharini: Good point, fixed.

sanjayprabhu (approved):

Issue #356, opened 2025-04-07
Add snapchain-protos as a standalone crate for protocol buffer definitions


christopherwxyz:
**Problem**

  We're currently refactoring Waypoint to use the latest Farcaster Snapchain proto definitions. However, our
  current approach requires copying the proto files directly from the Snapchain repository, which introduces several
  challenges:

  1. Maintenance burden: Each time Snapchain updates their proto files, we must manually copy them to our codebase
  2. Consistency risks: Different projects could end up using slightly different versions of the proto files
  3. Duplication: The proto files are copied across multiple projects, leading to duplication
  4. Versioning: Difficult to track which version of the proto files we're using

  **Proposed Solution**

  Create a dedicated `snapchain-protos` crate that:

  - Contains only the Protocol Buffer definitions
  - Is versioned separately from the main Snapchain codebase
  - Provides pre-compiled Rust types for the proto definitions
  - Follows semver for API compatibility

  **Benefits**

  1. Simplified integration: Projects like Waypoint can just add the crate as a dependency
  2. Version control: Clear versioning of the proto definitions
  3. Build performance: Pre-compiled proto files reduce build time
  4. Ecosystem health: Encourages consistent proto usage across the Farcaster ecosystem

  **Implementation**

  1. Create a new repository for `snapchain-protos`
  2. Add a GitHub Action to automatically sync proto files from the main Snapchain repo
  3. Publish to crates.io with clear versioning strategy
  4. Document breaking vs. non-breaking proto changes

  This approach is similar to how other projects handle their proto definitions (e.g., [gRPC-rs](https://github.com/tikv/grpc-rs/tree/master/proto)).

  **Effort**

  This would likely require 1-2 hours of initial setup, with minimal maintenance afterward as the sync process could be
  automated.

PR #358, opened 2025-04-07
refactor: add snapchain-proto as a dedicated crate for protocol buffer definitions

christopherwxyz:
Summary

  - Created a standalone snapchain-proto crate to contain all Protocol Buffer definitions
  - Moved proto files from src/proto/ to crates/proto/src/
  - Set up proper build configuration for proto generation
  - Updated imports throughout the codebase to use the new crate
  - Maintains identical interface through re-exports in the main lib.rs

  Benefits

  - Simplifies dependency management for external projects
  - Enables separate versioning for protocol definitions
  - Reduces code duplication across the ecosystem
  - Improves build performance by avoiding repeated proto compilation
  - Addresses requirements in #356


build.rs
```diff
@@ -1,29 +1,4 @@
...
-            "src/proto/admin_rpc.proto",
-            "src/proto/blocks.proto",
-            "src/proto/rpc.proto",
-            "src/proto/message.proto",
-            "src/proto/onchain_event.proto",
-            "src/proto/hub_event.proto",
-            "src/proto/username_proof.proto",
-            "src/proto/sync_trie.proto",
-            "src/proto/node_state.proto",
-            "src/proto/gossip.proto",
```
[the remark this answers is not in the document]
christopherwxyz: Done. 👍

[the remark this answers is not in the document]
sanjayprabhu: No worries. You can test locally by building the dockerfile locally, if it works there the action should also work.

crates/proto/Cargo.toml
```diff
@@ -0,0 +1,12 @@
+[package]
+name = "snapchain-proto"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+tonic = "0.12.3"
```
sanjayprabhu, line 7: Why do we need the new version? I'd try to keep the old version unless strictly required.

crates/proto/build.rs
```diff
@@ -0,0 +1,31 @@
+fn main() -> Result<(), Box<dyn std::error::Error>> {
+    let mut builder = tonic_build::configure();
+
+    // Custom type attributes required for malachite
+    builder = builder
+        .type_attribute("ShardHash", "#[derive(Eq, PartialOrd, Ord)]")
+        // Remove the Copy trait to avoid conflicts with prost's own Copy derive
```
sanjayprabhu, line 7: Why did this start failing now? Due to the version change? 


In this PR, I'd just only do the move and keep everything else the same as much as possible.

crates/proto/Cargo.toml
christopherwxyz: This is the same version as the parent crate?

https://github.com/farcasterxyz/snapchain/blob/main/Cargo.toml#L18

crates/proto/build.rs
christopherwxyz: Building locally causes a bunch of orphan implementation errors.

When I moved the proto definitions to their own crate, the main crate is still trying to implement methods directly on the proto types. So we can no longer add inherent impl blocks or implement foreign traits for those types.

It's turning into a non-trivial change, so will need to push this for now since mainnet launch is more important.

Issue #344, opened 2025-03-31
feat: Return block number as a field in Event


sanjayprabhu:
The block number is encoded in the event id, but it would be nice to have it returned directly as a usable field when queried via the Events API (both Subscribe and GetEvent). We don't need to store the block number separately, just extract the block number from the id and present it as a field.

Quazia:
I'd like to work on this! I believe Manan spoke to you about my interest in this issue.

sanjayprabhu:
Go for it, let me know if you have any questions

sanjayprabhu:
GetEvent needs to be implemented. If you want to do that separately or as part of this that would be helpful. No worries otherwise as well, we'll get to it.

Quazia:
I'll pick it up just let me know if you'd prefer it on the same PR or separate. Just let me know, thanks.

Quazia:
Similarly is it worth capturing an issue to implement `SubmitBulkMessages`?

sanjayprabhu:
Yes please.

PR #359, opened 2025-04-08
feat: Return block number as a field in Event #344

Quazia:
- Add a new `HubEventResponse` message to `request_response.proto`
- Add helper for converting a `HubEvent` to a `HubEventResponse` in `event.rs`
- Use helper in existing Events API endpoint (subscribe) in `server.rs`
- Change return type for Events API in `rpc.proto`
- Add tests to `server_tests`

### TODO

- [ ] Implement GetEvent with new `HubEventResponse`

### DIFF
--- a/src/network/server.rs
+++ b/src/network/server.rs
@@ -35,6 +35,7 @@ use crate::proto::UserNameProof;
 use crate::proto::UserNameType;
 use crate::proto::UsernameProofRequest;
 use crate::proto::UsernameProofsResponse;
+use crate::proto::HubEventResponse;
 use crate::proto::ValidationResponse;
 use crate::proto::VerificationAddAddressBody;
 use crate::proto::{Block, CastId, DbStats};
@@ -602,7 +603,7 @@ impl HubService for MyHubService {
         }))
     }
 
-    type SubscribeStream = ReceiverStream<Result<HubEvent, Status>>;
+    type SubscribeStream = ReceiverStream<Result<HubEventResponse, Status>>;
 
     async fn subscribe(
         &self,
@@ -611,7 +612,7 @@ impl HubService for MyHubService {
         authenticate_request(&request, &self.allowed_users)?;
 
         info!("Received call to [subscribe] RPC");
-        let (server_tx, client_rx) = mpsc::channel::<Result<HubEvent, Status>>(100);
+        let (server_tx, client_rx) = mpsc::channel::<Result<HubEventResponse, Status>>(100);
         let events_txs = match request.get_ref().shard_index {
             Some(shard_id) => match self.shard_senders.get(&(shard_id)) {
                 None => {
@@ -665,7 +666,8 @@ impl HubService for MyHubService {
                     for event in old_events.events {
                         if event_types.contains(&event.r#type) {
                             let event = Self::rewrite_hub_event(event);
-                            if let Err(_) = server_tx.send(Ok(event)).await {
+                            let response = HubEvent::create_hub_event_response(event);
+                            if let Err(_) = server_tx.send(Ok(response)).await {
                                 return;
                             }
                         }
@@ -692,7 +694,8 @@ impl HubService for MyHubService {
                             Ok(hub_event) => {
                                 if filtered_events.contains(&hub_event.r#type) {
                                     let hub_event = Self::rewrite_hub_event(hub_event);
-                                    match tx.send(Ok(hub_event)).await {
+                                    let response = HubEvent::create_hub_event_response(hub_event);
+                                    match tx.send(Ok(response)).await {
                                         Ok(_) => {}
                                         Err(_) => {
                                             // This means the client hung up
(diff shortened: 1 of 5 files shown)


src/proto/rpc.proto
```diff
@@ -23,8 +23,8 @@ service HubService {
   rpc GetFids(FidsRequest) returns (FidsResponse);
 
   // Events
-  rpc Subscribe(SubscribeRequest) returns (stream HubEvent);
-//  rpc GetEvent(EventRequest) returns (HubEvent);
+  rpc Subscribe(SubscribeRequest) returns (stream HubEventResponse);
```
manan19, line 26: this would break backwards compatibility for existing users of the `Subscribe`. Is there a path where we don't need to do a major version bump to make this change?

arthur-neynar: Yeah this is a breaking change. How widespread is current usage of snapchain/subscribe? We could add it to the `HubEvent` as optional but the `HubEvent` is passed around in more places than just the `Subscribe` return - relatively easy refactor if that's how we want to do it but I think switching all responses to Response objects while it's still easier to make these interface changes (not a lot of active clients) is the move. I think actually doing this for _all_ existing rpc function definitions makes sense and then it's just a single breaking change.

Quazia: Another option is to provide a `SubscribeEvent` fn with the new functionality and then deprecate the old `Subscribe` on the next major version bump?

manan19: @sanjayprabhu @aditiharini thoughts?^

sanjayprabhu: Hmm, why do we need to do this? I think we can just add a `block_number` field to HubEvent and keep the existing signature right?

arthur-neynar: I'll refactor to make that change - I think the main advantage or reason we would want a separate message is if whats stored for the HubEvent and what's returned through the endpoint continue to diverge it'd be easier to manage if they're separate.

So I will refactor to add an optional field to the `HubEvent` and then only populate that when providing it as a response.

Quazia: Ok I've refactored it to use the existing `HubEvent`

src/proto/hub_event.proto
```diff
@@ -59,4 +59,5 @@ message HubEvent {
     //    MergeStorageAdminRegistryEventBody merge_storage_admin_registry_event_body = 10;
     MergeOnChainEventBody merge_on_chain_event_body = 11;
   };
+  optional uint64 block_number = 12;
```
sanjayprabhu, line 62: Doesn't have to be optional, it will default to 0

src/network/server.rs
```diff
@@ -692,7 +693,8 @@ impl HubService for MyHubService {
                             Ok(hub_event) => {
                                 if filtered_events.contains(&hub_event.r#type) {
                                     let hub_event = Self::rewrite_hub_event(hub_event);
-                                    match tx.send(Ok(hub_event)).await {
+                                    let response = HubEvent::create_hub_event_response(hub_event);
```
sanjayprabhu, line 696: Let's just use the `rewrite_hub_event` above to add the block number

src/storage/store/account/event.rs
```diff
@@ -172,6 +172,25 @@ impl HubEvent {
...
     }
 
+    pub fn get_event(
+        db: Arc<RocksDB>,
+        event_id: u64,
+    ) -> Result<Option<HubEvent>, HubError> {
+        let key = Self::make_event_key(event_id);
+        let buf = db.get(&key)?;
+        if buf.is_none() {
+            return Ok(None); // Do we want to return none or an error?
```
sanjayprabhu, line 182: We should return a not found error

src/proto/hub_event.proto
Quazia: Addressed

src/network/server.rs
Quazia: Addressed

src/storage/store/account/event.rs
```diff
@@ -172,6 +172,25 @@ impl HubEvent {
...
+        event_id: u64,
+    ) -> Result<Option<HubEvent>, HubError> {
+        let key = Self::make_event_key(event_id);
+        let buf = db.get(&key)?;
+        if buf.is_none() {
+            return Err(HubError::not_found("Event not found"));
+        }
+
+        match HubEvent::decode(buf.unwrap().as_slice()) {
+            Ok(event) => Ok(Some(event)),
```
sanjayprabhu, line 186: Since we're returning error on failure, we can just use `Result<HubEvent, HubError>` instead of `Option<HubEvent>`

sanjayprabhu (commented):
Looks good, but looks like build is failing?

Quazia:
Sorry about that! Should be good now.

sanjayprabhu:
Build still failing? I'm surprised the precommit hook didn't catch it

Quazia:
Ok! Tests passing, build passing, everything is formatted. Hopefully there's nothing else I'm overlooking. Apologies again, will make sure to double-triple check each of these steps for future contributions.

Issue #357, opened 2025-04-07
http requests had their case change between hubble and snapchain


BlinkyStitt:
The docs for hubble have this example:

```
$ curl 'http://127.0.0.1:3381/v1/reactionById?fid=2&reaction_type=1&target_fid=1795&target_hash=0x7363f449bfb0e7f01c5a1cc0054768ed5146abc0'
```

But with snapchain, it fails with this error.

```
Invalid query parameters: missing field `reactionType`
```

1. I think snapchain is just missing a macro for handling the renames from snake case.

2. Also, `reaction_type` is no longer an int and now its the string name of the enum. So the example will need to be updated to match that.

3. The examples should also put quotes around the url. Mac's default shell now is zsh and that seems to get confused by the `?` in the url without quotes.

Quazia:
@sanjayprabhu Okay if I take this after #344 or you have something you prefer I start on?

sanjayprabhu:
Nope, go for it

sanjayprabhu:
Could you please swap everything to be `snake_case` to be consistent with hubs?

Quazia:
Looking into the existing hubble spec more I'm finding that route level query params are `snake_case` and component level query params are `camelCase` leading to some cases [like this](https://docs.neynar.com/reference/fetch-cast-reactions-1) where the existing query params in the hubble spec are mixed case. I can swap everything to be consistent with hubs, swap everything to be `snake_case` _or_ switch everything to `snake_case`/`camelCase` and alias anything that's inconsistent to ensure backwards compatibility in the short term. Let me know how to proceed 😊

Quazia:
**Digging deeper**
https://github.com/farcasterxyz/hub-monorepo/blob/9b0d11ccaf5e76b20bcca38ee06da889e7842cfe/protobufs/schemas/request_response.proto#L130-L137

Schema in hub monorepo looks consistent with Snapchain

Docs seem out of sync though
https://github.com/farcasterxyz/hub-monorepo/blob/9b0d11ccaf5e76b20bcca38ee06da889e7842cfe/apps/hubble/www/docs/docs/httpapi/reactions.md?plain=1#L11-L20

Quazia:
Is there a single canonical source of truth for the hub spec or is my best bet actually just looking through the hub code?

sanjayprabhu:
[...]

PR #363, opened 2025-04-09
fix: change query params to be backwards compatible with hubs

Quazia:
If this works as expected it'll be necesary to look over the hubble spec and the snapchain spec and alias any divergence.

Addresses #357

### DIFF
--- a/src/network/http_server.rs
+++ b/src/network/http_server.rs
@@ -297,11 +297,12 @@ pub struct CastsByParentRequest {
 #[derive(Debug, Clone, Deserialize, Serialize)]
 pub struct ReactionRequest {
     fid: u64,
-    #[serde(rename = "target_fid", skip_serializing_if = "Option::is_none")]
-    target_cast_id: Option<CastId>,
-    // I think this is actually a different type not just a rename - worth digging into more
-    #[serde(rename = "target_hash", skip_serializing_if = "Option::is_none")]
+    #[serde(skip_serializing_if = "Option::is_none")]
     target_url: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    target_fid: Option<u64>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    target_hash: Option<String>,
     reaction_type: ReactionType,
 }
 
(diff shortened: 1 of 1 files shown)


src/network/http_server.rs
```diff
@@ -308,11 +308,11 @@ pub struct ReactionRequest {
...
 pub struct ReactionsByFidRequest {
     fid: u64,
-    #[serde(rename = "reactionType")]
+    #[serde(rename = "reactionType", alias = "reaction_type")]
     reaction_type: ReactionType,
-    #[serde(rename = "pageSize", skip_serializing_if = "Option::is_none")]
+    #[serde(rename = "pageSize", alias = "page_size", skip_serializing_if = "Option::is_none")]
     page_size: Option<u32>,
-    #[serde(rename = "pageToken", skip_serializing_if = "Option::is_none")]
+    #[serde(rename = "pageToken", alias = "page_token", skip_serializing_if = "Option::is_none")]
```
sanjayprabhu, line 315: I think we can just get rid of `rename` and `alias` and it should pick up the default snake case?

@CassOnMars Any reason we have to use camel case? Snake case is what hubs use and it's breaking backwards compatibility

Quazia:
[I'm working off of this file for my source of truth](https://github.com/farcasterxyz/hub-monorepo/blob/main/apps/hubble/src/rpc/httpServer.ts#L202)

The biggest refactors were to [`/v1/reactionsByCast` and `/v1/reactionsByTarget`](https://github.com/farcasterxyz/snapchain/pull/363/files#diff-0f70fc9f1333a6b6b6ea261fb4f93593268ba3f64ab042182ec8160698c80478L322-R340)

A few endpoints don't expect params that are expected by the legacy hub server but that shouldn't be an issue they should just get ignored.

The biggest concern is around [`target_hash` vs `target_url` vs `targetUrl`](https://github.com/farcasterxyz/snapchain/pull/363/files#diff-0f70fc9f1333a6b6b6ea261fb4f93593268ba3f64ab042182ec8160698c80478R302-R304) as I believe they're actually different values not just different query names/cases.

Quazia:
For [/v1/reactionsByCast and /v1/reactionsByTarget](https://github.com/farcasterxyz/snapchain/pull/363/files#diff-0f70fc9f1333a6b6b6ea261fb4f93593268ba3f64ab042182ec8160698c80478L322-R340) I tried to make `reactionsByTarget` accept both query patterns and the nested `CastId` object since making that optional works there. For `reactionsByCast` I tried to make it less accepting. Having them overlap with the same request type like before didn't look like a viable option inclusive of backwards compatibility. Still unsure on the approach for [target_hash vs target_url vs targetUrl](https://github.com/farcasterxyz/snapchain/pull/363/files#diff-0f70fc9f1333a6b6b6ea261fb4f93593268ba3f64ab042182ec8160698c80478R302-R304) Maybe the best thing to do is add additional params for the legacy query params in that case since it just fully looks like those endpoints don't consume the hash info now. Maybe it's not a big deal if it's just ignored, I'm not familiar enough with the usage.

src/network/http_server.rs
```diff
@@ -297,18 +297,17 @@ pub struct CastsByParentRequest {
...
     fid: u64,
-    #[serde(rename = "reactionType")]
-    reaction_type: ReactionType,
-    #[serde(rename = "targetCastId", skip_serializing_if = "Option::is_none")]
+    #[serde(rename = "target_fid", skip_serializing_if = "Option::is_none")]
     target_cast_id: Option<CastId>,
-    #[serde(rename = "targetUrl", skip_serializing_if = "Option::is_none")]
+    // I think this is actually a different type not just a rename - worth digging into more
+    #[serde(rename = "target_hash", skip_serializing_if = "Option::is_none")]
     target_url: Option<String>,
```
sanjayprabhu, line 304: Yes, they are different. We should support both target_url and target_hash. Looks like hubs only support target_hash right now, so if you want to pick one, you can pick that and rename this field to `target_hash`

src/network/http_server.rs
```diff
@@ -297,18 +297,17 @@ pub struct CastsByParentRequest {
 #[derive(Debug, Clone, Deserialize, Serialize)]
 pub struct ReactionRequest {
     fid: u64,
-    #[serde(rename = "reactionType")]
-    reaction_type: ReactionType,
-    #[serde(rename = "targetCastId", skip_serializing_if = "Option::is_none")]
+    #[serde(rename = "target_fid", skip_serializing_if = "Option::is_none")]
     target_cast_id: Option<CastId>,
```
sanjayprabhu, line 301: If you're rename this target_fid, rename the field to match and have the data type be u64

src/network/http_server.rs
Quazia: I added support for both - that should still be backwards compatible and it sounds like this is more in line with desired functionality.

sanjayprabhu (approved):

PR #367, opened 2025-04-11
fix: retry connections to bootstrap peers

aditiharini:
If we're disconnected from any of the bootstrap peers, try to reconnect.

### DIFF
--- a/src/network/gossip.rs
+++ b/src/network/gossip.rs
@@ -310,10 +310,13 @@ impl SnapchainGossip {
     }
 
     pub async fn check_and_reconnect_to_bootstrap_peers(&mut self) {
-        warn!("Attempting to reconnect to bootstrap peers...");
-        for addr in &self.bootstrap_addrs {
-            if !self.connected_bootstrap_addrs.contains(addr) {
-                let _ = Self::dial(&mut self.swarm, &addr);
+        let connected_peers_count = self.swarm.connected_peers().count();
+        if !self.read_node || (self.read_node && connected_peers_count == 0) {
+            for addr in &self.bootstrap_addrs {
+                if !self.connected_bootstrap_addrs.contains(addr) {
+                    warn!("Attempting to reconnect to bootstrap peer: {}", addr);
+                    let _ = Self::dial(&mut self.swarm, &addr);
+                }
             }
         }
     }
--- a/src/network/gossip.rs
+++ b/src/network/gossip.rs
@@ -50,6 +50,7 @@ pub struct Config {
     pub announce_address: String,
     pub bootstrap_peers: String,
     pub contact_info_interval: Duration,
+    pub bootstrap_reconnect_interval: Duration,
 }
 
 impl Default for Config {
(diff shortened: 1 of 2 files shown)


src/network/gossip.rs
```diff
@@ -297,11 +310,9 @@ impl SnapchainGossip {
     }
 
     pub async fn check_and_reconnect_to_bootstrap_peers(&mut self) {
-        let connected_peers_count = self.swarm.connected_peers().count();
-        // If we have no connected peers, try to reconnect to bootstrap peers
-        if connected_peers_count == 0 && !self.bootstrap_addrs.is_empty() {
-            warn!("No connected peers. Attempting to reconnect to bootstrap peers...");
-            for addr in &self.bootstrap_addrs {
+        warn!("Attempting to reconnect to bootstrap peers...");
```
sanjayprabhu, line 313: Stray log line?

src/network/gossip.rs
```diff
@@ -297,11 +310,9 @@ impl SnapchainGossip {
...
+        for addr in &self.bootstrap_addrs {
+            if !self.connected_bootstrap_addrs.contains(addr) {
```
sanjayprabhu, line 315: We should add a log line here if we weren't actually connected

src/network/gossip.rs
```diff
@@ -297,11 +310,9 @@ impl SnapchainGossip {
     }
 
     pub async fn check_and_reconnect_to_bootstrap_peers(&mut self) {
-        let connected_peers_count = self.swarm.connected_peers().count();
-        // If we have no connected peers, try to reconnect to bootstrap peers
-        if connected_peers_count == 0 && !self.bootstrap_addrs.is_empty() {
```
sanjayprabhu, line 302: I think we should keep this behavior for read nodes? And only use the new one for validators. It's not required for read nodes to be connected to all validators

src/network/gossip.rs
```diff
@@ -271,11 +282,13 @@ impl SnapchainGossip {
...
             sync_channels: HashMap::new(),
             read_node,
-            bootstrap_addrs: config.bootstrap_addrs(),
+            bootstrap_addrs: config.bootstrap_addrs().into_iter().collect(),
             announce_address: config.announce_address.clone(),
             fc_network,
             contact_info_interval: config.contact_info_interval,
+            bootstrap_reconnect_interval: config.bootstrap_reconnect_interval,
             statsd_client,
+            connected_bootstrap_addrs: HashSet::new(),
```
sanjayprabhu, line 291: Do we need this? I think we can just call `self.swarm.connected_peers.contains`?

sanjayprabhu (approved):

src/network/gossip.rs
aditiharini: We can't use that because we don't have peer id in the config. We need to track peer ids for each bootstrap address in order to use connected_peers which will require similar logic.

sanjayprabhu: The peer id is derived from the public key, which we know ahead of time for validators. https://github.com/libp2p/rust-libp2p/blob/master/identity/src/peer_id.rs#L126-L130

Anyway, not a big deal. This works as well

PR #369, opened 2025-04-14
feat: Return error codes in standard format on submit

sanjayprabhu:
Return consistent error messages when submitting a message. Also return the error code as a separate field in the metadata

### DIFF
--- a/src/network/server.rs
+++ b/src/network/server.rs
@@ -49,22 +49,24 @@ use crate::storage::constants::OnChainEventPostfix;
 use crate::storage::constants::RootPrefix;
 use crate::storage::db::PageOptions;
 use crate::storage::db::RocksDbTransactionBatch;
-use crate::storage::store::account::message_bytes_decode;
 use crate::storage::store::account::MessagesPage;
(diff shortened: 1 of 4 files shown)


src/network/server.rs
```diff
@@ -420,14 +420,48 @@ impl HubService for MyHubService {
...
+                    Status::not_found(err.to_string())
+                } else if err_code.starts_with("db") || err_code.starts_with("internal") {
+                    Status::internal(err.to_string())
+                } else if err_code.starts_with("unavailable") {
+                    Status::unavailable(err.to_string())
+                } else {
+                    Status::unknown(err.to_string())
+                };
+                if let Ok(err_str) = AsciiMetadataValue::from_str(&err_code) {
+                    status.metadata_mut().insert("x-err-code", err_str);
```
sanjayprabhu, line 460: Hubble returns this directly as `errCode` but that's not valid with the rust library. I'll update hub-nodejs to read from this as well so clients get the error code populated correctly.

sanjayprabhu: https://github.com/farcasterxyz/hub-monorepo/pull/2567

Issue #373, opened 2025-04-15
missing grpc gzip compression


BlinkyStitt:
Hubble supports gzip compression on sends. When I try to connect to snapchain with code that works on hubble, it fails with:

```
Error: status: Unimplemented, message: "Content is compressed with `gzip` which isn't supported", details: [], metadata: MetadataMap { headers: {"content-type": "application/grpc", "grpc-accept-encoding": "identity", "date": "Tue, 15 Apr 2025 18:52:36 GMT", "content-length": "0"} }
```

This is the code that fails:

```
    let client = HubServiceClient::connect(gondor)
        .await?
        .accept_compressed(CompressionEncoding::Gzip)
        .send_compressed(CompressionEncoding::Gzip)
        .max_decoding_message_size(20 * 1024 * 1024)
        .max_encoding_message_size(20 * 1024 * 1024);
```

Removing "send_compressed" makes the code work. Was gzip support removed on purpose, or was it just missed?

sanjayprabhu:
Not intentional, likely differences in defaults between JS and rust GRPC libraries.

PR #385, opened 2025-04-17
feat: add gzip handling to the `hub_service`

Quazia:
Closses #373 

I don't know if we want to add a test for this - not sure how we'd go about it since we probably don't want to test `main.rs` directly.

Also noticed that the `tonic` and `tonic-build` versions are quite different. Worth updating them to be consistent or is there a reason for the older `tonic-build` version?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,7 +15,7 @@ serde = { version = "1.0", features = ["derive"] }
 libp2p-connection-limits = "0.5.0"
 serde_json = "1.0"
 sha2 = "0.10.6"
-tonic = "0.12.3"
+tonic = { version = "0.12.3", features = ["gzip"] }
 prost = "0.13.3"
 futures = "0.3.28"
 futures-core = "0.3.31"
--- a/src/main.rs
+++ b/src/main.rs
@@ -79,9 +79,8 @@ async fn start_servers(
         VERSION.unwrap_or("unknown").to_string(),
         gossip.swarm.local_peer_id().to_string(),
     ));
-    let grpc_service = service.clone();
     let grpc_shutdown_tx = shutdown_tx.clone();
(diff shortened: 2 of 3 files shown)


src/main.rs
```diff
@@ -78,11 +79,13 @@ async fn start_servers(
         VERSION.unwrap_or("unknown").to_string(),
         gossip.swarm.local_peer_id().to_string(),
     ));
-    let grpc_service = service.clone();
     let grpc_shutdown_tx = shutdown_tx.clone();
+    let hub_service = HubServiceServer::from_arc(service.clone())
+        .accept_compressed(CompressionEncoding::Gzip)
+        .send_compressed(CompressionEncoding::Gzip);
```
sanjayprabhu, line 85: Is this always or only if the header is present in the request? Just want to make sure it won't break existing clients.

Cargo.toml
```diff
@@ -15,7 +15,7 @@ serde = { version = "1.0", features = ["derive"] }
 libp2p-connection-limits = "0.5.0"
 serde_json = "1.0"
 sha2 = "0.10.6"
-tonic = "0.12.3"
+tonic = { version = "0.12.3", features = ["gzip"] }
```
sanjayprabhu, line 18: I think there were some issues upgrading tonic-build. Worth a shot if build passes, but if it requires a bunch of patches, probably not worth it.

BlinkyStitt: the automated tests pass. is there something more to check?

src/main.rs
BlinkyStitt: I'm 95% sure that it accepts it, but it doesn't require it.

Cargo.toml
Quazia: It requires a fair number of patches but still would be worth picking up at some point.

Issue #387, opened 2025-04-17
subscribing with from_id=None starts at the beginning instead of starting at now


BlinkyStitt:
Hubble used to start at "now" when you gave it no id. Snapchain should do the same

BlinkyStitt:
Workaround until this is fixed:

1. query the hub info
2. take the max height for the relevant shard and `<<14` it. use that as the from_id

PR #389, opened 2025-04-18
fix: start subscribe rpc from latest event if from_id is not specified

aditiharini:
We currently start from the earliest event when `fromId` is not specified. Hubs start from the latest and this is what we should do. 

https://github.com/farcasterxyz/snapchain/issues/387


cryptojcdenton:
Thank you for this!

Issue #382, opened 2025-04-16
bug: automatically detect and announce public address in contact info


sanjayprabhu:
ContactInfo currently announces 0.0.0.0 for all nodes unless an announce_address is manually configured. We should automatically detect the public address and use that if announce address is blank. Otherwise most nodes will not be able to connect with each other.

PR #402, opened 2025-04-18
feat: Use public IP as announce address

FrederikBolding:
Mirrors the implementation from hubs by fetching the public IP from `ipify.org` when no `announce_address` is specified.

Most of the tests rely on this not being the fact, so a couple of these had to be adjusted with proper configs to make CI pass.

Closes https://github.com/farcasterxyz/snapchain/issues/382

### DIFF
--- a/src/network/gossip.rs
+++ b/src/network/gossip.rs
@@ -25,6 +25,7 @@ use libp2p::{
 };
 use libp2p_connection_limits::ConnectionLimits;
 use prost::Message;
+use reqwest::Client;
 use serde::{Deserialize, Serialize};
 use std::collections::{HashMap, HashSet};
 use std::hash::{DefaultHasher, Hash, Hasher};
@@ -47,7 +48,7 @@ const CONTACT_INFO: &str = "contact-info";
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct Config {
     pub address: String,
-    pub announce_address: String,
+    pub announce_address: Option<String>,
     pub bootstrap_peers: String,
     pub contact_info_interval: Duration,
     pub bootstrap_reconnect_interval: Duration,
@@ -61,7 +62,7 @@ impl Default for Config {
         );
         Config {
             address: address.clone(),
-            announce_address: address,
+            announce_address: None,
             bootstrap_peers: "".to_string(),
             contact_info_interval: Duration::from_secs(300),
             bootstrap_reconnect_interval: Duration::from_secs(30),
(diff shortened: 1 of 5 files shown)


src/network/gossip.rs
```diff
@@ -47,7 +48,7 @@ const CONTACT_INFO: &str = "contact-info";
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct Config {
     pub address: String,
-    pub announce_address: String,
+    pub announce_address: Option<String>,
```
aditiharini, line 51: Can you leave this as a `String` and can we have empty string mean unspecified? In our config, we define envvars for config parameters that are different by host and use that in the config. It's more annoying to omit a config parameter than to set the envvar for it to empty string.

aditiharini:
To confirm, did you test this by running a node locally? Did you see that announce address was properly detected?

src/network/gossip.rs
```diff
@@ -292,6 +302,31 @@ impl SnapchainGossip {
...
+            return address.clone();
+        }
+
+        // If no config-defined announce IP exists, detect the public IP.
+        // Falling back to the address also defined in the config
+        let public_ip = Self::get_public_ip().await;
+
+        match public_ip {
+            Ok(address) => format!("/ip4/{}/udp/{}/quic-v1", address, DEFAULT_GOSSIP_PORT),
+            Err(_) => config.address.clone(),
```
aditiharini, line 316: log the error here

src/network/gossip.rs
```diff
@@ -273,6 +281,8 @@ impl SnapchainGossip {
         // Listen on all assigned port for this id
         swarm.listen_on(config.address.parse()?)?;
 
+        let announce_address = Self::get_announce_address(config).await;
```
aditiharini, line 284: Let's log as info what we're using as the announce_address

FrederikBolding:
> To confirm, did you test this by running a node locally? Did you see that announce address was properly detected?

Yep, I did!

src/network/gossip.rs
FrederikBolding: Addressed! https://github.com/farcasterxyz/snapchain/pull/402/commits/5641a6c019b50324da2946c46a17a03ea1b9dcba

src/network/gossip.rs
FrederikBolding: Addressed! https://github.com/farcasterxyz/snapchain/pull/402/commits/5641a6c019b50324da2946c46a17a03ea1b9dcba

src/network/gossip.rs
FrederikBolding: Addressed! https://github.com/farcasterxyz/snapchain/pull/402/commits/5641a6c019b50324da2946c46a17a03ea1b9dcba

PR #404, opened 2025-04-19
fix docker logs command

alongcamewayne:
Updated the command on the getting started page to match the actual service name

CleanShot 2025-04-19 at 16 36 46@2x


alongcamewayne:
i realized this is probably because of my directory structure. sorry!

Issue #397, opened 2025-04-18
event timestamp and shard id missing in snapchain's events


BlinkyStitt:
With hubble, the event ids are the timestamp the hub received the event and a 12 bit sequence number. 

With snapchain, the event ids are a block height and a 14 bit index number. 

Sanjay said that the hub events will have the shard id and receive time in them soon. I didn't see a task about it, so I opened this one so that we can track it easily. this should maybe me two issues, but i didn't want to spam.

sanjayprabhu:
The shardId has been implemented but won't show up in the nodeJS client until we do this https://github.com/farcasterxyz/snapchain/issues/383

The timestamp change turned out to be more complicated. We'll do it next week. Thanks for filing this.

PR #409, opened 2025-04-21
feat: add timestamp to hub event

aditiharini:
This makes it much easier to determine how far behind you are. 

https://github.com/farcasterxyz/snapchain/issues/397

### DIFF
--- a/src/network/server.rs
+++ b/src/network/server.rs
@@ -900,6 +900,7 @@ impl HubService for MyHubService {
                 if let Ok(Some(chunk)) = chunk {
                     hub_event.produced_at = chunk.header.as_ref().unwrap().timestamp
                 }
+
                 Ok(Response::new(hub_event))
             }
             Err(err) => Err(Status::internal(err.to_string())),
--- a/src/network/server.rs
+++ b/src/network/server.rs
@@ -6,7 +6,6 @@ use crate::core::validations;
 use crate::core::validations::verification::VerificationAddressClaim;
 use crate::mempool::mempool::{MempoolRequest, MempoolSource};
 use crate::mempool::routing;
-use crate::proto;
 use crate::proto::cast_add_body;
 use crate::proto::casts_by_parent_request;
 use crate::proto::hub_service_server::HubService;
@@ -38,6 +37,7 @@ use crate::proto::UsernameProofRequest;
 use crate::proto::UsernameProofsResponse;
 use crate::proto::ValidationResponse;
 use crate::proto::VerificationAddAddressBody;
+use crate::proto::{self, Height};
 use crate::proto::{Block, CastId, DbStats};
 use crate::proto::{
     BlocksRequest, EventRequest, EventsRequest, EventsResponse, ShardChunksRequest,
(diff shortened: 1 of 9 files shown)


src/network/server.rs
```diff
@@ -737,7 +738,36 @@ impl HubService for MyHubService {
...
+                                            shard_index: store.shard_id,
+                                            block_number: event.block_number,
+                                        }
+                                        .as_u64(),
+                                    );
+                                    last_chunk = chunk.unwrap_or(None);
+                                }
+
+                                if let Some(chunk) = &last_chunk {
+                                    event.produced_at = chunk.header.as_ref().unwrap().timestamp;
```
sanjayprabhu, line 768: Hmm, I think this should go into rewrite_hub_event. It's best to have a single place where the events are mutated. Otherwise we're going to miss things.

src/storage/store/engine_tests.rs
```diff
@@ -878,9 +880,11 @@ mod tests {
         assert_eq!(stored_onchain_events.len(), 1);
 
         // Hub events are generated
-        let events = HubEvent::get_events(engine.db.clone(), 0, None, None).unwrap();
+        let mut events = HubEvent::get_events(engine.db.clone(), 0, None, None).unwrap();
         assert_eq!(1, events.events.len());
-        let received_event = event_rx.recv().await.unwrap();
+        let mut received_event = event_rx.recv().await.unwrap();
+        received_event.produced_at = 0;
+        events.events.get_mut(0).unwrap().produced_at = 0;
```
sanjayprabhu, line 887: Why do we need to set to 0?

src/network/server_tests.rs
```diff
@@ -546,6 +549,95 @@ mod tests {
...
+                    .header
+                    .as_ref()
+                    .unwrap()
+                    .height
+                    .unwrap()
+                    .block_number,
+                0,
+            ),
+        });
+        add_auth_header(&mut req, USER_NAME, PASSWORD);
```
sanjayprabhu, line 633: We should remove the auth requirement for this endpoint

src/storage/store/engine_tests.rs
aditiharini: The assert compares the events with each other exactly. We use current timestamp for populating shard chunk timestamp so this timestamp field differs. 

It's kind of annoying to thread through an optional or default timestamp-- we have to modify the commit_shard_chunk function which is used a lot.

aditiharini: added a comment -- the timestamp is not stored in the db but it is populated in the generated event based on the timestamp in the shard chunk. that time is now current time.

sanjayprabhu (approved):

src/network/server.rs
```diff
@@ -415,10 +415,17 @@ impl MyHubService {
...
+        mut hub_event: HubEvent,
+        shard_index: u32,
+        timestamp: Option<u64>,
+    ) -> HubEvent {
         let (block_number, _) = HubEventIdGenerator::extract_height_and_seq(hub_event.id);
         hub_event.block_number = block_number;
         hub_event.shard_index = shard_index;
+        if let Some(timestamp) = timestamp {
+            hub_event.timestamp = timestamp;
+        }
```
sanjayprabhu, line 428: We can delete this logic 3 days after releasing this right? Worth adding a todo.

sanjayprabhu (approved):

src/network/server.rs
aditiharini: This logic is because the events we emit live already have timestamp populated

aditiharini: We don't write timestamp to disk right now

Issue #406, opened 2025-04-21
`onChainSignersByFid` returns different response on Snapchain vs. Hubs


Quazia:
There seems to be a discrepancy in the return type of the event

[This causes issues when trying to validate in the miniapps `createVerifyAppKeyWithHub` function
](https://github.com/farcasterxyz/miniapps/blob/362251dcd6efdf7ffaeaa8089a7d39ac16140616/packages/frame-node/src/farcaster.ts#L41-L82) as [reported here
](https://warpcast.com/hellno.eth/0x33c39051)

## Snapchain return
```curl -X GET "http://localhost:3381/v1/onChainSignersByFid?fid=12345" -H "Content-Type: application/json"```

```
{
    "events": [
        {
            "type": "EventTypeSigner",
            "chainId": 10,
            "blockNumber": 111900036,
            "blockHash": [36, 17, 124, 11, 251, 230, 248, 64, 51, 25, 139, 135, 249, 45, 137, 106, 184, 163, 63, 234, 54, 172, 120, 20, 64, 206, 125, 22, 36, 234, 158, 228],
            "blockTimestamp": 1699398849,
            "transactionHash": [68, 215, 95, 212, 14, 53, 93, 84, 130, 138, 184, 19, 226, 160, 240, 111, 61, 19, 236, 73, 113, 48, 243, 182, 149, 244, 66, 57, 69, 171, 243, 131],
            "logIndex": 18,
            "fid": 12345,
            "signerEventBody": {
                "key": [0, 176, 131, 123, 233, 78, 233, 226, 197, 244, 38, 59, 164, 177, 216, 49, 148, 236, 46, 170, 218, 100, 71, 96, 3, 5, 171, 182, 241, 154, 50, 41],
                "keyType": 1,
                "eventType": "Add",
[...]

PR #410, opened 2025-04-21
refactor: change expected return types for `SignerEventBody`

Quazia:
Closes https://github.com/farcasterxyz/snapchain/issues/406

### DIFF
--- a/src/network/http_server.rs
+++ b/src/network/http_server.rs
@@ -574,8 +574,8 @@ pub enum IdRegisterEventType {
 
 #[derive(Debug, Clone, Deserialize, Serialize)]
 pub struct SignerEventBody {
-    #[serde(with = "serdebase64")]
-    pub key: Vec<u8>, // Hex or base?
+    #[serde(with = "serdehex")]
+    pub key: Vec<u8>,
     #[serde(rename = "keyType")]
     pub key_type: u32,
     #[serde(rename = "eventType")]
--- a/src/network/http_server.rs
+++ b/src/network/http_server.rs
@@ -546,22 +546,22 @@ pub struct UserNameProof {
 pub struct UsernameProofsResponse {
     pub proofs: Vec<UserNameProof>,
 }
-
+#[allow(non_camel_case_types)]
 #[derive(Debug, Clone, Deserialize, Serialize)]
 pub enum OnChainEventType {
-    EventTypeNone = 0,
-    EventTypeSigner = 1,
-    EventTypeSignerMigrated = 2,
-    EventTypeIdRegister = 3,
-    EventTypeStorageRent = 4,
+    EVENT_TYPE_NONE = 0,
+    EVENT_TYPE_SIGNER = 1,
+    EVENT_TYPE_SIGNER_MIGRATED = 2,
+    EVENT_TYPE_ID_REGISTER = 3,
+    EVENT_TYPE_STORAGE_RENT = 4,
 }
-
+#[allow(non_camel_case_types)]
 #[derive(Debug, Clone, Deserialize, Serialize)]
 pub enum SignerEventType {
-    None = 0,
-    Add = 1,
-    Remove = 2,
-    AdminReset = 3,
+    SIGNER_EVENT_TYPE_NONE = 0,
+    SIGNER_EVENT_TYPE_ADD = 1,
+    SIGNER_EVENT_TYPE_REMOVE = 2,
+    SIGNER_EVENT_TYPE_ADMIN_RESET = 3,
 }
 
 #[derive(Debug, Clone, Deserialize, Serialize)]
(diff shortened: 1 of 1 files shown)


src/network/http_server.rs
```diff
@@ -1887,11 +1919,11 @@ impl HubHttpService for HubHttpServiceImpl {
                     }
                     return OnChainEvent {
                         r#type: match e.r#type {
-                            1 => OnChainEventType::EventTypeSigner,
-                            2 => OnChainEventType::EventTypeSignerMigrated,
-                            3 => OnChainEventType::EventTypeIdRegister,
-                            4 => OnChainEventType::EventTypeStorageRent,
-                            _ => OnChainEventType::EventTypeNone,
+                            1 => "SIGNER".to_string(),
```
aditiharini, line 1922: Is this to avoid the EVENT_TYPE prefix you'd get if you use the as_str... function?

Quazia: Yeah - is there a better way to be handling this?

src/network/http_server.rs
```diff
@@ -44,6 +44,27 @@ mod serdebase64 {
...
+    use serde::{de::Error, Deserializer, Serializer};
+    /// Serialize a byte vector to a "0x"‑prefixed hex string.
+    pub fn serialize<S: Serializer>(v: &Vec<u8>, s: S) -> Result<S::Ok, S::Error> {
+        // hex::encode turns &[u8] → lowercase hex (no prefix)
+        let mut prefixed = String::with_capacity(v.len() * 2 + 2);
+        prefixed.push_str("0x");
+        prefixed.push_str(&hex::encode(v));
+        s.serialize_str(&prefixed)
+    }
+    /// Deserialize from either a "0x"‑prefixed or raw hex string into Vec<u8>.
```
aditiharini, line 59: Why is 0x optional? Is this behavior taken from hubs?

Quazia: Hubs seemed to be returning them _without_ the prefix so thought it was best to handle either

src/network/http_server.rs
```diff
@@ -471,14 +492,16 @@ pub enum IdRegisterEventType {
 
 #[derive(Debug, Clone, Deserialize, Serialize)]
 pub struct SignerEventBody {
-    pub key: Vec<u8>,
+    #[serde(with = "serdebase64")]
```
aditiharini, line 495: based on hub code i think this should be hex

Quazia: Ok thanks - making the change right now

src/network/http_server.rs
aditiharini: I think this is fine. in all the places where you're doing these manual conversions can you leave a note about the fact that we're doing this for compatibility. it's a bit odd so would be good to justify.

aditiharini:
Were you able to test this with a local node?

src/network/http_server.rs
Quazia: Sure and yeah agreed. Another option would be to do the conversion then strip off the prefix but that seems more complex.  I'll add the comments at least.

Quazia:
> > Were you able to test this with a local node?
> 
> ```
> curl -X GET "http://localhost:3381/v1/onChainSignersByFid?fid=12345" -H "Content-Type: application/json"
> {
>   "events": [
>     {
>       "type": "SIGNER",
>       "chainId": 10,
>       "blockNumber": 111900036,
>       "blockHash": "0x24117c0bfbe6f84033198b87f92d896ab8a33fea36ac781440ce7d1624ea9ee4",
>       "blockTimestamp": 1699398849,
>       "transactionHash": "0x44d75fd40e355d54828ab813e2a0f06f3d13ec497130f3b695f4423945abf383",
>       "logIndex": 18,
>       "fid": 12345,
>       "signerEventBody": {
>         "key": "0x00b0837be94ee9e2c5f4263ba4b1d83194ec2eaada6447600305abb6f19a3229",
>         "keyType": 1,
>         "eventType": "ADD",
>         "metadata": "AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAACAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAjwAAAAAAAAAAAAAAAAALveQ3XmTo1/YR8BT7drpQNBVWWAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAIAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAZO5/LAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAABBXXVNpfxqvN4K8N7TMq4TQZx5ic4clkfqNDSc34pfuWcyW+ItefdepWrsbt0IMIDSfD2VslbLtJFB8koUD6gugRsAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA",
>         "metadataType": "NONE"
>       },
>       "txIndex": 5,
>       "version": 0
>     }
>   ]
> }
> ```

Some of the type strings still don't look right but everything else looks right let me fix that.

PR #411, opened 2025-04-21
fix: implement /v1/events HTTP API endpoint

### LINKED ISSUE #395: bug: implement /v1/events http api
We have this on hubs and we should have it on snapchain too.

gr4yha7:
Closes #395

### DIFF
--- a/src/network/http_server.rs
+++ b/src/network/http_server.rs
@@ -18,8 +18,7 @@ use crate::proto::{
     UserDataType, UserNameType,
 };
 use crate::proto::{
-    casts_by_parent_request, link_request, links_by_target_request, on_chain_event,
-    reaction_request, reactions_by_target_request, LinksByFidRequest, Protocol,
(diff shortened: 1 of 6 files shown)


src/network/http_server.rs
```diff
@@ -671,6 +670,137 @@ pub struct OnChainEventRequest {
     reverse: Option<bool>,
 }
 
+#[derive(Debug, Clone, Deserialize, Serialize)]
+pub enum HubEventType {
```
Quazia, line 674: [Can we keep these consistent with hubs?](https://github.com/farcasterxyz/hub-monorepo/blob/be0e240eb9ebd3bd66e8751d6789096ed57242a7/protobufs/schemas/hub_event.proto#L7)

Quazia (commented):
It looks like some of the return types are going to be slightly different with how you have it implemented.

Can you make sure that the shape of the return matches with the [existing hubble implementaton](https://lamia.farcaster.xyz:2281/v1/events)?

sanjayprabhu:
Thank you, cleaned up a little bit and merged in https://github.com/farcasterxyz/snapchain/pull/451

PR #421, opened 2025-04-22
fix: return empty `nextPageToken` for `PagedResponse`

Quazia:
Previously an empty string was returned for all hubble paged responses, this keeps that consistently in the response with an empty string.


Quazia:
@aditiharini  Assuming the sync issues are fixed could I get a review on this today please?

sanjayprabhu (approved):

PR #427, opened 2025-04-23
fix: disconnect unexpected peers from read nodes

aditiharini:
Read nodes are timing out on requests to non-bootstrap peers. Kick any peer that dials a read node because this is unexpected.


sanjayprabhu:
Let's hold off on this for now.

PR #432, opened 2025-04-23
fix: adjust sync reply logs

aditiharini:
Fix the sync response log and add timings.


sanjayprabhu:
Do we still need this?

PR #435, opened 2025-04-24
fix: have sync actor and consensus actor use separate hosts

aditiharini:


sanjayprabhu:
Do we still need this?

PR #439, opened 2025-04-24
feat: handle slow peers in gossip

aditiharini:
Blacklist peers with more than 3 timeouts in the past hour. Each hour, reset the timeout count and allow reconnecting to the peer. 

Also added a `snapchain_version` field to contact info to allow detecting nodes that haven't upgraded.

### DIFF
--- a/src/network/gossip.rs
+++ b/src/network/gossip.rs
@@ -393,13 +393,22 @@ impl SnapchainGossip {
         num_timeouts > 3
     }
 
-    fn maybe_disconnect_slow_peer(&mut self, peer_id: PeerId) {
+    fn maybe_disconnect_slow_peer(&mut self, peer_id: PeerId, source: &str) {
         if self.is_peer_slow(&peer_id) && !self.bootstrap_peer_ids.contains(&peer_id) {
+            let is_connected = self.swarm.is_connected(&peer_id);
             info!(
                 peer_id = peer_id.to_string(),
+                source = source,
+                is_connected = is_connected,
                 "Disconnecting peer due to too many timeouts"
             );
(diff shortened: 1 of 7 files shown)


src/network/gossip.rs
```diff
@@ -293,6 +298,11 @@ impl SnapchainGossip {
 
         info!("Using {} as announce address", announce_address);
 
+        let peer_cache = CacheBuilder::new(100)
+            .time_to_live(Duration::from_secs(60 * 60))
```
sanjayprabhu, line 302: Let's make both of these constants. Max capacity should be higher, like 2K

src/network/gossip.rs
```diff
@@ -374,13 +385,26 @@ impl SnapchainGossip {
...
 
+    pub fn update_slow_peers(&mut self, peer_id: &PeerId) {
+        let num_timeouts = self.slow_peers.get(peer_id).unwrap_or(0);
+        if num_timeouts <= 3 {
+            self.swarm
+                .behaviour_mut()
+                .gossipsub
+                .remove_blacklisted_peer(peer_id);
+        } else {
+            self.swarm.behaviour_mut().gossipsub.blacklist_peer(peer_id);
```
sanjayprabhu, line 396: I don't think blackslisting affects the connection (only rejects the messages), so this may not do what we want. I think we need to disconnect manually.

src/network/gossip.rs
```diff
@@ -614,6 +639,8 @@ impl SnapchainGossip {
             return;
         }
 
+        self.update_slow_peers(&peer_id);
+
         if self.enable_autodiscovery {
```
sanjayprabhu, line 644: Let's also turn this on by default

Issue #438, opened 2025-04-24
fix: Same fnames are associated with multiple fids


sanjayprabhu:
```
curl https://snapchain.uno.fun:3381/v1/userNameProofsByFid\?fid\=20270
{"proofs":[{"timestamp":1739048246,"name":"heavygweit.eth","owner":"0x18fc3c60a80ba26a076cf633c239df0e2e80a797","signature":"pRC0LB7grqPBRxUNYfBk06PNzSk506+okTs1HaQabbpnZqf7x/rDxzzTbzXcMvZyVSga1ShR1EwjzY376Q1iE
xs=","fid":20270,"type":"USERNAME_TYPE_ENS_L1"},{"timestamp":1741384226,"name":"erica","owner":"0x92ce59c18a97646e9a7e011653d8417d3a08bb2b","signature":"AMNgHFFe3/4gjnEo9H+Jwvt7jgvqr2FRWDBd3wKBinFnmo5wYlA75ZoZ0
kG9C0c5ajwpTPr9DVR42xroJJRjvRw=","fid":20270,"type":"USERNAME_TYPE_FNAME"}]}%

curl https://snapchain.uno.fun:3381/v1/userNameProofsByFid\?fid\=211428
{"proofs":[{"timestamp":1703050844,"name":"erica","owner":"0x2b4d92e7626c5fc56cb4641f6f758563de1f6bdc","signature":"BQtC/dp7CnMJofuKLLyaX0u/JBrsdPUxkflmXZufVy1PRSrIB5Ea97aYAhlILW9/2n+Z8jqxnJYbRwG5k0+i+Rs=","fid
":211428,"type":"USERNAME_TYPE_FNAME"}]}%
```

Hubs have a similar issue (but fid is consistent)
```
curl https://andil.merkle.zone:2281/v1/userNameProofsByFid\?fid\=211428
{"proofs":[{"timestamp":1741384226,"name":"erica","owner":"0x92ce59c18a97646e9a7e011653d8417d3a08bb2b","signature":"AMNgHFFe3/4gjnEo9H+Jwvt7jgvqr2FRWDBd3wKBinFnmo5wYlA75ZoZ0kG9C0c5ajwpTPr9DVR42xroJJRjvRw=","fid":20270,"type":"USERNAME_TYPE_FNAME"}]}%

curl https://andil.merkle.zone:2281/v1/userNameProofsByFid\?fid\=20270
[...]

PR #442, opened 2025-04-24
fix: Prevent issues with fnames being associated with multiple fids

FrederikBolding:
Fixes two problems, when an fname was transferred the secondary index was not correctly deleted causing outdated info to still be returned by the API. In addition, fname transfers are now mirrored to both the sender and received shard to prevent outdated info in one of the shards during a transfer.

Fixes https://github.com/farcasterxyz/snapchain/issues/438

### DIFF
--- a/src/mempool/mempool.rs
+++ b/src/mempool/mempool.rs
@@ -488,6 +488,34 @@ impl Mempool {
             .message_router
             .route_fid(fid, self.read_node_mempool.num_shards);
 
+        self.insert_into_shard(shard_id, message.clone(), source.clone())
+            .await;
+
+        // Fname transfers are mirrored to both the sender and receiver shard.
+        match message {
+            MempoolMessage::ValidatorMessage(ref inner_message) => {
+                if let Some(fname_transfer) = &inner_message.fname_transfer {
+                    let mirror_shard_id = self
+                        .read_node_mempool
+                        .message_router
+                        .route_fid(fname_transfer.from_fid, self.read_node_mempool.num_shards);
+
+                    if mirror_shard_id != shard_id {
+                        self.insert_into_shard(mirror_shard_id, message, source)
+                            .await;
+                    }
+                }
+            }
+            _ => {}
+        }
+    }
+
+    async fn insert_into_shard(
+        &mut self,
+        shard_id: u32,
+        message: MempoolMessage,
+        source: MempoolSource,
+    ) {
         match self.messages.get_mut(&shard_id) {
             Some(shard_messages) => {
                 if shard_messages.contains_key(&message.mempool_key()) {
--- a/src/mempool/mempool.rs
+++ b/src/mempool/mempool.rs
@@ -492,21 +492,18 @@ impl Mempool {
             .await;
 
         // Fname transfers are mirrored to both the sender and receiver shard.
-        match message {
-            MempoolMessage::ValidatorMessage(ref inner_message) => {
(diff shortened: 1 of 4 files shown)


tests/consensus_test.rs
```diff
@@ -770,3 +784,93 @@ async fn test_read_node() {
...
     }
 }
+
+#[tokio::test]
+#[serial]
+async fn test_cross_shard_interactions() {
+    let env_filter = EnvFilter::try_from_default_env().unwrap_or_else(|_| EnvFilter::new("warn"));
+    let _ = tracing_subscriber::fmt()
+        .with_env_filter(env_filter)
+        .try_init();
```
sanjayprabhu, line 794: Delete? If we need logs for debugging there's an `test_helper::enable_logging`

src/storage/store/account/name_registry_events.rs
```diff
@@ -81,6 +82,12 @@ pub fn put_username_proof_transaction(
 
     let secondary_key = make_fname_username_proof_by_fid_key(username_proof.fid);
     txn.put(secondary_key, primary_key);
+
+    // If a username is being transferred, remove the existing secondary key.
+    if let Some(existing_fid) = existing_fid {
+        let secondary_key = make_fname_username_proof_by_fid_key(existing_fid);
+        txn.delete(secondary_key);
```
sanjayprabhu, line 89: Hmm, have to consider how to handle existing bad data. This doesn't break consensus since it's just an index change. We might need a migration to fix, but we don't have any migration harness code. Will ticket this, let me know if you're interested in taking that on.

src/mempool/mempool.rs
```diff
@@ -488,6 +488,31 @@ impl Mempool {
...
+        // Fname transfers are mirrored to both the sender and receiver shard.
+        if let MempoolMessage::ValidatorMessage(inner_message) = &message {
+            if let Some(fname_transfer) = &inner_message.fname_transfer {
+                let mirror_shard_id = self
+                    .read_node_mempool
+                    .message_router
+                    .route_fid(fname_transfer.from_fid, self.read_node_mempool.num_shards);
+
+                if mirror_shard_id != shard_id {
+                    self.insert_into_shard(mirror_shard_id, message, source)
```
sanjayprabhu, line 503: hmm, this technically affects consensus. But we might be ok if nodes upgrade at different times since this logic is only hit in the proposer? Other nodes will behave the same regardless of version, as long as the transfer is in the transactions list.

sanjayprabhu (commented):
This is a tricky change, we should be careful about merging this. @aditiharini is working on some logic to do protocol version upgrades. We might want to wait until we have that in place first.

tests/consensus_test.rs
FrederikBolding: Ah! Sorry copied that from another test, fixed!

src/mempool/mempool.rs
FrederikBolding: Not sure it will affect consensus in practice since it only changes where the message is stored tbh.

src/storage/store/account/name_registry_events.rs
FrederikBolding: Feels like it's probably something to deal with outside of this PR

sanjayprabhu:
Updated to put this behind a protocol version check here https://github.com/farcasterxyz/snapchain/pull/522

Thank you!

PR #453, opened 2025-04-29
fix: Populate message data in events

sanjayprabhu:
When sending events back to the client, ensure `message.data` is populated, if only `dataBytes` was present when the message was submitted


sanjayprabhu:
Hmm, actually, looks like events persist messages with message.data populated, unlike stores. So this is working right now. Feels better to have this be more explicit like this, but this code will not be hit right now. Let me think on this.

PR #458, opened 2025-04-30
feat: set up version schedule

aditiharini:
Prepares for protocol upgrades.

### DIFF
--- a/src/consensus/proposer.rs
+++ b/src/consensus/proposer.rs
@@ -19,7 +19,35 @@ use tokio::time::Instant;
 use tokio::{select, time};
 use tracing::{error, warn};
 
-pub const PROTOCOL_VERSION: u32 = 1;
+#[derive(Eq, PartialEq, Clone, Copy)]
+pub enum ProtocolVersion {
+    UNKNOWN = 0,
+    ONE = 1,
+    TWO = 2,
+}
+
+struct VersionSchedule {
+    version: ProtocolVersion,
+    active_at: u64,
+}
+
+const PROTOCOL_VERSION_SCHEDULE: [VersionSchedule; 1] = [VersionSchedule {
+    version: ProtocolVersion::ONE,
+    active_at: 0,
+}];
+
+pub fn get_version_at_timestamp(timestamp: u64) -> ProtocolVersion {
+    PROTOCOL_VERSION_SCHEDULE
+        .iter()
+        .filter(|version| version.active_at <= timestamp)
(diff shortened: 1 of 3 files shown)


src/consensus/proposer.rs
```diff
@@ -19,7 +19,37 @@ use tokio::time::Instant;
...
+#[derive(Eq, PartialEq, Clone, Copy)]
+pub enum ProtocolVersion {
+    UNKNOWN = 0,
+    ONE = 1,
+    TWO = 2,
+}
+
+struct VersionSchedule {
+    version: ProtocolVersion,
+    active_at: u64,
```
sanjayprabhu, line 33: Worth adding a comment about what kind of timestamp this is (seconds since unix epoch?)

src/consensus/proposer.rs
```diff
@@ -19,7 +19,37 @@ use tokio::time::Instant;
...
+}
+
+const PROTOCOL_VERSION_SCHEDULE: [VersionSchedule; 1] = [VersionSchedule {
+    version: ProtocolVersion::ONE,
+    active_at: 0,
+}];
+
+pub fn get_version_at_timestamp(timestamp: u64) -> ProtocolVersion {
```
sanjayprabhu, line 41: Probably on your list, but we should add tests for this

src/consensus/read_validator.rs
```diff
@@ -44,7 +45,22 @@ impl ReadValidator {
...
     }
 
+    fn check_version(value: &DecidedValue) {
+        let proposal = value.value.as_ref().unwrap();
+        match proposal {
+            proto::decided_value::Value::Block(block) => {
+                let header = block.header.as_ref().unwrap();
+                if header.version > get_current_version() as u32 {
+                    // Maybe we want to upgrade this to a panic
+                    error!("Node should be upgraded now, received a block with version too high. Version {}.", header.version)
```
sanjayprabhu, line 55: Yeah, I think we should log and then panic

src/consensus/proposer.rs
```diff
@@ -495,7 +534,11 @@ impl Proposer for BlockProposer {
         proposal
     }
 
-    fn add_proposed_value(&mut self, full_proposal: &FullProposal) -> Validity {
+    fn add_proposed_value(
+        &mut self,
+        full_proposal: &FullProposal,
+        proposal_source: ProposalSource,
```
sanjayprabhu, line 540: Why do we need `proposal_source`?

PR #463, opened 2025-05-02
fix: Read camelCase pageOption in http requests for backwards compatibility

sanjayprabhu:
Report https://warpcast.com/nikolaii.eth/0x2de48ca6

Hubble was reading certain params in snake case and certain params in camelCase. We standardized on snake_case in snapchain, but this broke frequently used pagination params which were in camelCase in hubble.

So, read both snake case and camel case to preserve backwards compatibility. Prioritizes snake_case.


src/network/http_server.rs
```diff
@@ -1702,16 +1890,7 @@ impl HubHttpService for HubHttpServiceImpl {
...
-                FidTimestampRequest {
-                    fid: req.fid,
-                    page_size: req.page_size,
-                    page_token: req.page_token,
-                    reverse: req.reverse,
-                    start_timestamp: req.start_timestamp,
-                    stop_timestamp: req.stop_timestamp,
-                },
-            ))
+            .get_casts_by_fid(tonic::Request::<proto::FidRequest>::new(req.to_proto()))
```
sanjayprabhu, line 1893: This is a hubble bug. All other endpoints call the respective get_<type>_by_fid method, not get_all_<type> which also returns removes. Fixed to be consistent.

sanjayprabhu: Hmm, actually, people are relying on this behavior. Will revert to keep existing functionality.

aditiharini (approved):

PR #465, opened 2025-05-05
Adds Primary Address support to Snapchain

PirosB3:
Implementation should mirror [the updated FIP](https://github.com/farcasterxyz/protocol/discussions/141)

This is an initial draft of the PR, I have a hunch that it's ~80% there but it's also worth opening up for quick feedback to make sure I don't invest time in the wrong direction.


src/storage/store/engine.rs
```diff
@@ -826,11 +927,23 @@ impl ShardEngine {
                 .user_data_store
                 .merge(msg, txn_batch)
                 .map_err(|e| MessageValidationError::StoreError(e)),
-            MessageType::VerificationAddEthAddress | MessageType::VerificationRemove => self
+            MessageType::VerificationAddEthAddress => self
```
PirosB3, line 930: @aditiharini one small note - There is no mention here of `VerificationAddEthSolana` when there should be. Am I wrong or is there a bug?

src/core/validations/message.rs
```diff
@@ -262,6 +264,41 @@ fn validate_longitude(value: &str) -> Result<(), ValidationError> {
...
+        return Ok(());
+    }
+
+    if !input.starts_with("0x") || input.len() != 42 {
+        return Err(ValidationError::InvalidDataLength);
+    }
+
+    let parsed = Address::from_hex(input).map_err(|_| ValidationError::InvalidData)?;
+
+    let checksummed = parsed.to_checksum(None);
```
PirosB3, line 279: I noticed there are some better helper functions to validate Sol/Eth addresses. I'll swap my code out with those, where appropriate

src/storage/store/engine.rs
```diff
@@ -826,11 +927,23 @@ impl ShardEngine {
...
                 .map_err(|e| MessageValidationError::StoreError(e)),
+            MessageType::VerificationRemove => {
+                let result = self
+                    .stores
+                    .verification_store
+                    .merge(msg, txn_batch)
+                    .map_err(|e| MessageValidationError::StoreError(e));
+
+                if let Ok(hub_event) = result.as_ref() {
+                    self.handle_verification_merge_hooks(msg.fid(), hub_event, txn_batch)?;
```
aditiharini, line 943: Instead otf calling `handle_verification_merge_hooks` here, I would do it one level higher. I would look at `hub_event.deleted_messages`, iterate over each of those and if any of them is a verification add then call `handle_verification_merge_hooks` on the add. This ensures that we're responding to a deleted verification add message.

src/storage/store/engine.rs
```diff
@@ -814,6 +830,118 @@ impl ShardEngine {
         Ok((account_root, events, validation_errors))
     }
 
+    /// Checks if a removed verification was set as the user's primary address and revokes it if necessary.
+    ///
+    /// When a verification is removed from a user's account, this function ensures that if the
+    /// removed address was also set as their primary address (for either Ethereum or Solana),
+    /// the primary address setting is also revoked to maintain consistency.
+    ///
+    fn check_and_revoke_primary_address(
```
PirosB3, line 839: @sanjayprabhu to review please

src/core/validations/message.rs
```diff
@@ -424,6 +461,12 @@ pub fn validate_user_data_add_body(body: &UserDataBody) -> Result<(), Validation
         UserDataType::Github => {
             validate_github_username(&body.value)?;
         }
+        UserDataType::UserDataPrimaryAddressEthereum => {
+            validate_user_data_primary_address_ethereum(&body.value)?;
+        }
+        UserDataType::UserDataPrimaryAddressSolana => {
+            validate_user_data_primary_address_solana(&body.value)?;
```
sanjayprabhu, line 468: There's a `validate_eth_address` and `validate_sol_address` being used by the verification validation. We should use the same functions. Inconsistencies would causes issues where a verification that was accepted wouldn't allowed to be used as the primary address

src/proto/message.proto
```diff
@@ -102,6 +102,8 @@ enum UserDataType {
   USER_DATA_TYPE_LOCATION = 7; // Current location for the user
   USER_DATA_TYPE_TWITTER = 8; // Username of user on twitter
   USER_DATA_TYPE_GITHUB = 9; // Username of user on github
+  USER_DATA_PRIMARY_ADDRESS_ETHEREUM = 10; // Primary address for the user on Ethereum
+  USER_DATA_PRIMARY_ADDRESS_SOLANA = 11; // Primary address for the user on Solana
```
sanjayprabhu, line 106: FYI, this is going to conflict with @aditiharini's pro changes for a banner image. One of you will need to update the sequence number

src/storage/store/engine.rs
```diff
@@ -1080,6 +1210,12 @@ impl ShardEngine {
                 if user_data.r#type == proto::UserDataType::Username as i32 {
                     self.validate_username(message_data.fid, &user_data.value, txn_batch)?;
                 }
+                if user_data.r#type == proto::UserDataType::UserDataPrimaryAddressEthereum as i32 {
+                    self.validate_ethereum_address_ownership(message_data.fid, &user_data.value)?;
+                }
+                if user_data.r#type == proto::UserDataType::UserDataPrimaryAddressSolana as i32 {
+                    self.validate_solana_address_ownership(message_data.fid, &user_data.value)?;
```
sanjayprabhu, line 1217: I would have these call `verify_fid_owns_address` directly with the the appropriate protocol. Don't convert the address, just convert it to bytes and just check that it matches the verification exactly.

src/storage/store/engine.rs
```diff
@@ -1474,6 +1644,55 @@ impl ShardEngine {
...
+            if let Some(msg_data) = &message.data {
+                if let Some(Body::VerificationAddAddressBody(verification)) = &msg_data.body {
+                    return verification.protocol == target_protocol
+                        && verification.address == target_address;
+                }
+            }
+            false
+        };
+
+        let page_result = self.stores.verification_store.get_adds_by_fid(
```
sanjayprabhu, line 1677: Use `get_verification_add` which takes in an address. It's uses an address index, it's better than paginating. And then once you get the verification you can validate the protocol matches.

src/storage/store/engine.rs
```diff
@@ -813,6 +831,118 @@ impl ShardEngine {
...
+        if user_data_body.value == parsed_address {
+            self.stores.user_data_store.revoke(&user_data, txn_batch)?;
+        }
+
+        Ok(())
+    }
+
+    /// Takes a list of merge events of type VerificationAddAddress and revokes the user's primary address
+    /// if it was deleted as part of a verification remove message.
+    fn revoke_primary_addresses_for_deleted_verifications(
```
sanjayprabhu, line 901: Let's call this function `handle_delete_side_effects`

src/storage/store/engine.rs
```diff
@@ -813,6 +831,118 @@ impl ShardEngine {
...
+        // the user's primary address if it was deleted as part of a verification remove message.
+        for event in events {
+            if let Some(hub_event::Body::MergeMessageBody(merge_body)) = &event.body {
+                for deleted_message in &merge_body.deleted_messages {
+                    // Check if this was a verification add message that was deleted
+                    if let Some(data) = &deleted_message.data {
+                        // NOTE: VerificationAddEthAddress is used for both Ethereum and Solana verifications.
+                        // we differentiate between the two by checking the protocol field.
+                        if data.r#type == MessageType::VerificationAddEthAddress as i32 {
+                            if let Some(Body::VerificationAddAddressBody(body)) = &data.body {
```
sanjayprabhu, line 923: I think pattern matching is more concise here. Something like
```rust
       let data = deleted_message.data.as_ref().ok_or(MessageValidationError::NoMessageData)?
       match &data.body {
            Some(proto::message_data::Body::VerificationAddAddressBody(user_data)) => {
```

Also makes it easier to add more branches in the future

src/storage/store/engine.rs
```diff
@@ -813,6 +831,118 @@ impl ShardEngine {
...
+            _ => return Err(MessageValidationError::AddressNotPartOfVerification),
+        };
+
+        // Parse address based on protocol
+        let parsed_address = match protocol {
+            proto::Protocol::Ethereum => {
+                Address::from_slice(&deleted_verification.address).to_checksum(None)
+            }
+            proto::Protocol::Solana => bs58::encode(&deleted_verification.address).into_string(),
+        };
```
sanjayprabhu, line 859: Don't parse addresses, just treat them as bytes

src/storage/store/engine.rs
```diff
@@ -813,6 +831,118 @@ impl ShardEngine {
...
+        // Check if the stored primary address matches the one being removed
+        let data_ref = match &user_data.data {
+            Some(data) => data,
+            None => return Ok(()),
+        };
+
+        let user_data_body = match &data_ref.body {
+            Some(Body::UserDataBody(body)) => body,
+            _ => return Ok(()),
+        };
```
sanjayprabhu, line 889: Would be nice to condense these three matches somehow, but not a big deal. I don't think it's unreasonable to call unwrap since these should be valid messages.

src/storage/store/engine.rs
```diff
@@ -813,6 +831,118 @@ impl ShardEngine {
...
+
+        // If the primary address matches the one being removed, revoke it
+        if user_data_body.value == parsed_address {
+            self.stores.user_data_store.revoke(&user_data, txn_batch)?;
```
sanjayprabhu, line 893: So, if you modify the store, we need to update the trie and emit the events. Otherwise the account state in the db will not reflect the merkle trie state.

We should capture the return value here and then return the list of events from `handle_delete_side_effects`, then similar to the `prune_messages` call, call update_trie for each event and then add these to the global list of events being returned from `replay_snapchain_txn`

src/storage/store/engine_tests.rs
```diff
@@ -576,6 +577,335 @@ mod tests {
...
+    }
+
+    #[tokio::test]
+    async fn test_primary_address_revoked_when_verification_deleted() {
+        let timestamp = messages_factory::farcaster_time();
+        let (mut engine, _tmpdir) = test_helper::new_engine_with_options(EngineOptions {
+            limits: None,
+            db: None,
+            messages_request_tx: None,
+            network: Some(proto::FarcasterNetwork::Devnet), // Enables all protocol features including primary addresses
```
sanjayprabhu, line 662: I don't think you need this, it defaults to devnet

src/storage/store/engine_tests.rs
```diff
@@ -576,6 +577,335 @@ mod tests {
...
+        let verification_remove = messages_factory::verifications::create_verification_remove(
+            FID3_FOR_TEST,
+            eth_address_bytes.clone(),
+            Some(timestamp + 2),
+            None,
+        );
+        commit_message(&mut engine, &verification_remove).await;
+
+        // Step 4: Verify the primary address was automatically revoked
+        let user_data_result_after = engine.get_user_data_by_fid_and_type(
```
sanjayprabhu, line 729: Also assert that the userdata add does not exist in the trie anymore via `test_helper::message_exists_in_trie`

src/storage/store/engine_tests.rs
```diff
@@ -576,6 +577,335 @@ mod tests {
...
+            test_helper::default_custody_address(),
+            &mut engine,
+        )
+        .await;
+
+        // Try to set a primary address without having a verification for it
+        let eth_address = "1234567890abcdef1234567890abcdef12345678";
+        let eth_address_bytes = hex::decode(eth_address).unwrap();
+        let address_instance = alloy_primitives::Address::from_slice(&eth_address_bytes);
+        let eth_address_checksummed = address_instance.to_checksum(None);
```
sanjayprabhu, line 770: Do we need to do all this? Can we just use `hex::decode`?

src/storage/store/engine.rs
PirosB3: I don't think this would work if I called `verify_fid_owns_address`  directly because doing so I would be comparing the bytes of an encoded address (`"0x00000...".bytes()`) with the bytes of a decoded address, right?

src/storage/store/engine.rs
PirosB3: Just like in the earlier case, the UserDataAdd message contains a checksum-encoded address. So if I don’t decode the bytes into a proper address here, I wouldn’t be comparing equivalent values.

From spec
```
enum UserDataType {
  USER_DATA_PRIMARY_ADDRESS_ETHEREUM = 7; 
  // Primary Ethereum Address
  // e.g., "0xd5596099ec95b32ddC3F22814785a253f6a09D56"

  USER_DATA_PRIMARY_ADDRESS_SOLANA = 8; 
  // Primary Solana Address
  // e.g., "4TciSRW38RGNiTSKmQamQvxUg4epWKBirBvG8LCh3ahZ"
}
```

src/core/validations/message.rs
PirosB3: changed!

src/storage/store/engine.rs
PirosB3: - user_data_body.value.as_bytes() - this gives us the UTF-8 bytes of the checksummed string 
  - deleted_verification.address - this is the raw address bytes

PirosB3: Sample test: https://github.com/farcasterxyz/snapchain/pull/465/files#diff-6524ecb933c89497277e430bf140343f1a2d597c2001b24993ff1dd60cbcffc2R697

src/core/validations/message.rs
```diff
@@ -424,6 +462,12 @@ pub fn validate_user_data_add_body(body: &UserDataBody) -> Result<(), Validation
         UserDataType::Github => {
             validate_github_username(&body.value)?;
         }
+        UserDataType::UserDataPrimaryAddressEthereum => {
+            validate_user_data_primary_address_ethereum(&body.value)?;
+        }
+        UserDataType::UserDataPrimaryAddressSolana => {
```
sanjayprabhu, line 468: Please add the check for `ProtocolFeature::PrimaryAddresses` here so the messages will be rejected if they are submitted before the version activates. Please add a test for it as well.

PirosB3: done

sanjayprabhu (approved):
Looks good, nice work

PR #471, opened 2025-05-08
fix: `getidregistryonchaineventbyaddress`

### LINKED ISSUE #470: fix: `GetIdRegistryOnChainEventByAddress`
There's an issue occuring when we call the `GetIdRegistryOnChainEventByAddress` grpc endpoint right now with the same syntax we did previously on hubs.

[I believe it's due to this](https://github.com/farcasterxyz/snapchain/blob/d6d3f11568e3937d7572857a06711d6cdc5fa682/src/network/server.rs#L1786)

So I'm thinking we just implement that, we probably also want to add it to the `http_server` as a follow on.

Quazia:
Implements `get_id_registry_on_chain_event_by_address`

I fought with how to handle the `get_onchain_events` we could alternatively make the `fid` on that optional but I didn't want to mess with the existing interfaces too much. I also initially added an iterator on the events but backtracked on that since it didn't seem right but you can see if in the previous commits. I ended up adding a `get_all_onchain_events` which doesn't take an FID at all. If we'd prefer to make the `fid` optional on the existing function I can do that.

I added a test and it's passing but we might want to put the iterator through the ringer a little more. Also callout I did not test this with complete production state so I don't know how much this will effect latency on a cold get before the cache spins up so that might be worth testing but I don't have the full snapchain local right now (waiting on a drive to get that running again)

@aditiharini 

Closes #470

### DIFF
--- a/src/network/server_tests.rs
+++ b/src/network/server_tests.rs
@@ -1258,10 +1258,9 @@ mod tests {
         )
         .await;
 
-        let mut request = Request::new(proto::IdRegistryEventByAddressRequest {
+        let request = Request::new(proto::IdRegistryEventByAddressRequest {
             address: owner.clone(),
         });
-        add_auth_header(&mut request, USER_NAME, PASSWORD);
         let response = service
             .get_id_registry_on_chain_event_by_address(request)
             .await
--- a/src/network/server_tests.rs
+++ b/src/network/server_tests.rs
@@ -1243,4 +1243,34 @@ mod tests {
 
         assert_eq!(shard2_ref.next_page_token, vec![110, 117, 108, 108].into());
     }
+
+    #[tokio::test]
+    async fn test_get_id_registry_event_by_address() {
+        let (_stores, _senders, [mut engine1, _], service) = make_server(None).await;
+        let owner = test_helper::default_custody_address();
+        let fid = SHARD1_FID;
+        // Should we write a bunch of users to test the iteration or is this sufficient?
+        test_helper::register_user(
+            fid,
+            test_helper::default_signer(),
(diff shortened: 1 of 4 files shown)


src/storage/store/account/onchain_event_store.rs
```diff
@@ -425,6 +425,34 @@ impl OnchainEventStore {
         Ok(onchain_events)
     }
 
+    pub fn get_all_onchain_events(
```
aditiharini, line 428: Let's make fid optional on the existing `get_onchain_events` function.

aditiharini:
Do you want to add the http api too?

src/network/server_tests.rs
```diff
@@ -1243,4 +1243,34 @@ mod tests {
...
+            test_helper::default_signer(),
+            owner.clone(),
+            &mut engine1,
+        )
+        .await;
+
+        let mut request = Request::new(proto::IdRegistryEventByAddressRequest {
+            address: owner.clone(),
+        });
+        add_auth_header(&mut request, USER_NAME, PASSWORD);
```
aditiharini, line 1264: do you need auth for this endpoint?

Quazia: Nope - removed

src/storage/store/account/onchain_event_store.rs
Quazia: Ok done-zo 😁

Quazia:
> Do you want to add the http api too?

I'm happy to add these but can I do it in another PR just to get this out?

aditiharini (approved):

PR #490, opened 2025-05-15
first pass at adding trusted peers

BlinkyStitt:
this doesn't do anything about making them unbannable or some of the other thing's I around maximum peer tracking.

but it at least dials them often. I think that's a good start.

also, i noticed the Self::dial connects without 'unknown_peer_id'. Shouldn't it be using the id from inside the Multiaddr? or at least checking their id when they reply? the current behavior seems unsafe unless i'm missing some checks elsewhere.

### DIFF
--- a/src/network/gossip.rs
+++ b/src/network/gossip.rs
@@ -51,6 +51,7 @@ pub struct Config {
     pub address: String,
     pub announce_address: String,
     pub bootstrap_peers: String,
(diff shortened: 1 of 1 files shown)


BlinkyStitt:
I think maybe dialing bootstraps, then listening, then dialing trusted peers would be a better order. I'm not sure. Happy to hear suggestions.

PR #500, opened 2025-05-21
feat: messages in the mempool should be persisted on graceful shutdown

Quazia:
Initially the persistence was entirely in mempool which was a little cleaner but required passing in the shard db instance so I refactored. I think we probably want to be persisting on message vs on shutdown but that's more error prone and this is a good stop-gap in the near term.

Also wondering if instead of `insert_message_for_restore` we want to pass the messages into the mempool on init?


sanjayprabhu:
Closing for now, not currently high priority. We can revisit later.

PR #507, opened 2025-05-28
perf: Fix perf test script

aditiharini:
- Adds implementation to Admin APIs (`submit_user_name_proof`, `submit_on_chain_event`) for perf tests. These are only accessible on unspecified or dev networks.
- Updates perftest w/ a sleep to allow for previous messages to be processed
- Updates CLI utils to utilize auth for `send_message` and `send_on_chain_event`


sanjayprabhu (approved):
Looks good!

PR #515, opened 2025-06-03
feat: Support Basenames

sanjayprabhu:
Implement https://github.com/farcasterxyz/protocol/discussions/234 to support basenames

Todo:
 - [x] Unit Tests
 - [x] Username length validation update
 - [ ] Manual verification


src/connectors/onchain_events/mod.rs
```diff
@@ -151,7 +191,7 @@ impl RealL1Client {
 }
 
 #[async_trait]
-impl L1Client for RealL1Client {
+impl ChainAPI for RealL1Client {
     async fn resolve_ens_name(&self, name: String) -> Result<Address, EnsError> {
         foundry_common::ens::NameOrAddress::Name(name)
```
aditiharini, line 196: Does this work? I thought that this hard codes the address for the ens resolver on eth. Does the basename resolver use the same address?

sanjayprabhu: Yeah, you're right. Fixed by bringing in the foundry validation code here.

src/network/server.rs
```diff
@@ -302,98 +307,109 @@ impl MyHubService {
...
-                            format!("ENS resolution error: {}", err.to_string()).as_str(),
-                        )
-                    })?
-                    .to_vec();
-
-                if resolved_ens_address != proof.owner {
+                self.chain_clients.for_chain(Chain::EthMainnet)?
+            }
+            Ok(UserNameType::UsernameTypeBasename) => {
+                if !name.ends_with(".eth") {
```
aditiharini, line 383: .base.eth?

src/network/server.rs
```diff
@@ -302,98 +307,109 @@ impl MyHubService {
...
+                HubError::validation_failure(
+                    format!("ENS resolution error: {}", err.to_string()).as_str(),
+                )
+            })?
+            .to_vec();
+
+        Ok(resolved_ens_address)
     }
 
     async fn validate_ens_username(&self, fid: u64, name: String) -> Result<(), HubError> {
```
aditiharini, line 415: This function only queries for username proofs with `UsernameTypeEnsL1`. Not sure how the tests are passing but I think we also need to incorporate the basename type.

src/network/server.rs
sanjayprabhu: Fixed. We can probably get rid of this check though, by the time we get here message validation would've already ensured this is correct.

PR #516, opened 2025-06-03
feat: support pro users

aditiharini:
Add support for pro users.


src/storage/store/account/onchain_event_store.rs
```diff
@@ -553,6 +555,43 @@ impl OnchainEventStore {
         }
         Ok(None)
     }
+    pub fn is_tier_subscription_active_at(
```
sanjayprabhu, line 558: I think we'll need to add caching for this, but doesn't have to happen in this PR. Let's add a todo.

src/storage/store/engine_tests.rs
```diff
@@ -1864,4 +1865,100 @@ mod tests {
             false
         ); // Revoked
     }
+
+    #[tokio::test]
+    async fn pro_users_get_10k_casts() {
```
sanjayprabhu, line 1870: Should also add a test that this fails for non pro users, and for pro users before the protocol feature activates

src/network/http_server.rs
```diff
@@ -1818,6 +1834,16 @@ fn map_proto_on_chain_event_to_json_on_chain_event(
                 expiry: body.expiry,
             });
         }
+        Some(on_chain_event::Body::TierPurchaseEventBody(body)) => {
```
sanjayprabhu, line 1837: We should also update the user limits endpoint to return whether a user is pro or not. So we have a single place to get user status

src/connectors/onchain_events/mod.rs
```diff
@@ -52,19 +54,30 @@ sol!(
...
+
+pub static STORAGE_REGISTRY: Address = address!("00000000fcce7f938e7ae6d3c335bd6a1a7c593d");
+
+pub static KEY_REGISTRY: Address = address!("00000000Fc1237824fb747aBDE0FF18990E59b7e");
 
-static KEY_REGISTRY: Address = address!("00000000Fc1237824fb747aBDE0FF18990E59b7e");
+pub static ID_REGISTRY: Address = address!("00000000Fc6c5F01Fc30151999387Bb99A9f489b");
 
-static ID_REGISTRY: Address = address!("00000000Fc6c5F01Fc30151999387Bb99A9f489b");
+pub static TIER_REGISTRY: Address = address!("6F22091496694c370E5EbEF861c8f69472fF89B9");
```
aditiharini, line 70: Need to put the prod one in here

src/connectors/onchain_events/mod.rs
```diff
@@ -52,19 +54,30 @@ sol!(
...
 // Note these are the registry addresses, not the resolver addresses. We look up the resolver from the registry.
 static ETH_L1_ENS_REGISTRY: Address = address!("00000000000C2E074eC69A0dFb2997BA6C7d2e1e");
 static BASE_MAINNET_ENS_REGISTRY: Address = address!("0xB94704422c2a1E396835A571837Aa5AE53285a95");
 
 // For reference, in case it needs to be specified manually
-const FIRST_BLOCK: u64 = 108864739;
-static CHAIN_ID: u32 = 10; // OP mainnet
+pub const OP_MAINNET_FIRST_BLOCK: u64 = 108864739;
+pub static OP_MAINNET_CHAIN_ID: u32 = 10; // OP mainnet
+pub const BASE_MAINNET_FIRST_BLOCK: u64 = 31108599;
```
aditiharini, line 79: Update to block number for prod contract

PR #520, opened 2025-06-05
Refactor and update consensus tests

defnotpaul:
- Adds some basic assertions to tests to ensure sync / consensus is working
- Fixes port assignment issue so tests can be run concurrently
- Introduces some higher-level helper methods and assertions for writing tests

### DIFF
--- a/tests/consensus_test.rs
+++ b/tests/consensus_test.rs
@@ -834,7 +834,11 @@ async fn test_basic_consensus() {
         .await
         .unwrap();
 
-    assert_network_has_num_blocks(&network, 1);
+    // Wait for nodes to reach the next block height
+    let next_block_height = network.max_block_height() + 1;
+    network.wait_for_block(next_block_height).await;
+
+    assert_network_has_num_blocks(&network, next_block_height);
     assert_network_has_num_shard_chunks(&network, 3);
     assert_network_has_messages(&network, 1);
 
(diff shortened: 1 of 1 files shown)


tests/consensus_test.rs
```diff
@@ -501,348 +576,385 @@ impl TestNetwork {
...
+    network.register_and_wait_for_fid(1000).await;
+    let cast = network
+        .send_and_wait_for_cast(1000, "Hello, world")
+        .await
+        .unwrap();
 
-    let mut network = TestNetwork::create(4, num_shards, 3220).await;
+    assert_network_has_num_blocks(&network, 1);
+    assert_network_has_num_shard_chunks(&network, 3);
+    assert_network_has_messages(&network, 1);
```
aditiharini, line 839: Let's remove this assertion given we've added the one below?

tests/consensus_test.rs
```diff
@@ -501,348 +576,385 @@ impl TestNetwork {
...
+    network.start_validators().await;
+
+    network.register_and_wait_for_fid(1000).await;
+    let cast = network
+        .send_and_wait_for_cast(1000, "Hello, world")
+        .await
+        .unwrap();
 
-    let mut network = TestNetwork::create(4, num_shards, 3220).await;
+    assert_network_has_num_blocks(&network, 1);
```
aditiharini, line 837: Were we previously asserting 3 here? Did the test fail with 3?

aditiharini (approved):

tests/consensus_test.rs
defnotpaul: I've adjusted this test to have all the nodes sync at the next block height before running assertions, and then ensuring that they've all reached that block. 

I think one of the issues that caused flakiness in the previous tests was that the nodes were all potentially at different heights, and therefore we relied on sleeping/producing blocks in order for them to converge. The approach I'm favoring now is to be more explicit about sync through these of these `wait*` methods to ensure nodes have converged.

Issue #529, opened 2025-06-10
feat: validating addresses


christopherwxyz:
## Problem
Currently, when authenticating with SIWF (Sign-In With Farcaster) in mini apps, we need to check if an address is valid as either:
1. A custody address (from IdRegistry)
2. An auth address (from KeyRegistry with keyType=2)
Since it's not obvious which registry the address comes from, we have to race both checks, which is inefficient and adds unnecessary complexity.

## Current Workaround
We currently use `GetOnChainSignersByFid` which returns all signer types, but:
- It doesn't filter by keyType
- It includes regular signers (keyType=1) which we don't need for auth
- We still need to separately check IdRegistry for custody addresses

### Create a new dedicated endpoint
- Something like `GetFidAddressType` which checks both registries + verified addresses
- This endpoint would check:
  - IdRegistry for custody addresses
  - KeyRegistry for auth addresses (keyType=2)
  - Verified addresses for the FID
- Return a unified response indicating the address capabilities (since verified addresses could also be auth addresses)
- Example: `GetFidAddressType(fid: number, address: string) -> { isCustody: boolean, isAuth: boolean, isVerified: boolean, valid: boolean }`

## Benefits
- Reduces API calls and latency
- Simplifies client implementation
- Makes the auth flow more straightforward for mini app developers
- Clearer separation between auth addresses and regular signers
- Supports verified addresses in addition to custody and auth addresses

## Use Case
This is particularly important for mini apps using SIWF where users authenticate with their Farcaster identity using either their custody address or an authorized auth address. Today farcaster.xyz uses custody addresses in mobile, but an auth address on web.

sanjayprabhu:
I like `ValidateAddressForFid`, maybe it's something like `GetFidAddressAssociation(fid, address) -> { type: 'custody' | 'auth' | 'verified' | null }` and that way it can also check for verified addresses.

christopherwxyz:
Updated. There's a possibility a verified address could be an auth address, as well.

PR #535, opened 2025-06-11
feat: unified fid address type endpoint

christopherwxyz:
Add GetFidAddressType endpoint for unified address validation as reference in https://github.com/farcasterxyz/snapchain/issues/529

## Summary
Adds a new `GetFidAddressType` endpoint that checks if an address is valid as a custody address, auth address, or verified address in a single API call.

## Problem
Currently, SIWF (Sign-In With Farcaster) authentication requires racing multiple API calls to check if an address is valid:
- Check IdRegistry for custody addresses
- Check KeyRegistry for auth addresses (keyType=2)
- Check verified addresses

This is inefficient and adds unnecessary complexity.

## Solution
New endpoint `GetFidAddressType(fid, address)` returns:
```proto
{
  is_custody: boolean,   // Address owns the FID (IdRegistry)
  is_auth: boolean,      // Address is an auth key (KeyRegistry keyType=2)
  is_verified: boolean,  // Address is verified for this FID
  valid: boolean         // Any of the above is true
}
```

## Changes
- Added proto definitions for `FidAddressTypeRequest` and `FidAddressTypeResponse`
- Implemented gRPC endpoint in `server.rs`
- Added HTTP endpoint `/v1/fidAddressType`
- Added comprehensive tests

## Benefits
- Single API call instead of multiple racing requests
- Simplified client implementation for SIWF
- Clear distinction between address types
- Supports farcaster.xyz using custody addresses on mobile and auth addresses on web

### DIFF
--- a/src/network/server.rs
+++ b/src/network/server.rs
@@ -1894,13 +1894,10 @@ impl HubService for MyHubService {
             }
         }
 
-        let valid = is_custody || is_auth || is_verified;
-
         Ok(Response::new(FidAddressTypeResponse {
             is_custody,
             is_auth,
             is_verified,
-            valid,
         }))
     }
 
--- a/src/network/server.rs
+++ b/src/network/server.rs
@@ -26,6 +26,7 @@ use crate::proto::OnChainEventRequest;
 use crate::proto::OnChainEventResponse;
 use crate::proto::ReactionType;
 use crate::proto::ReactionsByTargetRequest;
+use crate::proto::SignerEventType;
 use crate::proto::SignerRequest;
 use crate::proto::TrieNodeMetadataRequest;
 use crate::proto::TrieNodeMetadataResponse;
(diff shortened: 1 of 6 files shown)


src/network/server.rs
```diff
@@ -1836,6 +1838,72 @@ impl HubService for MyHubService {
...
+            }
+        }
+
+        let valid = is_custody || is_auth || is_verified;
+
+        Ok(Response::new(FidAddressTypeResponse {
+            is_custody,
+            is_auth,
+            is_verified,
+            valid,
```
sanjayprabhu, line 1903: Do we need `valid` if we're breaking down the individual types?

sanjayprabhu (commented):
Thank you, looks great. Just one question.

src/network/server.rs
christopherwxyz: Nope.

PR #541, opened 2025-06-15
feat: Add endpoint for connected peers

FrederikBolding:
Add RPC endpoint for getting the currently connected peers. To do this, introduces a `BTreeMap` in `SnapchainGossip` that stores the contact info received from other peers.

### DIFF
--- a/src/network/gossip.rs
+++ b/src/network/gossip.rs
@@ -28,7 +28,7 @@ use libp2p_connection_limits::ConnectionLimits;
 use prost::Message;
 use reqwest::Client;
 use serde::{Deserialize, Serialize};
-use std::collections::{HashMap, HashSet};
+use std::collections::{BTreeMap, HashMap, HashSet};
 use std::hash::{DefaultHasher, Hash, Hasher};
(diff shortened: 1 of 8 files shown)


src/network/gossip.rs
```diff
@@ -588,6 +591,10 @@ impl SnapchainGossip {
         );
 
         let contact_peer_id = PeerId::from_bytes(&contact_info_body.peer_id).unwrap();
+
+        self.peers
```
FrederikBolding, line 595: @sanjayprabhu Seemingly `handle_contact_info` is not used for validator nodes, is there another preferred place to store peer information?

sanjayprabhu: Hmm, I think we should still insert into the map for validators. The intent behind the check was that the validators will never use autodiscovery and don't need to initiate connection requests to other peers.

sanjayprabhu (commented):
Nice, looks good.

sanjayprabhu (approved):
Were you able to test manually?

PR #542, opened 2025-06-16
constants.proto to define protocol constants

vrypan:
When using protocol messages, I always have to check and redefine FARCASTER_EPOCH, sometimes in multiple modules in my code.

This PR introduces an idiomatic way to define such constants and make them available to any protobuf consumer (`CONST_FARCASTER_EPOCH` in this case).

Other constants, such as cast size, longcast size, etc could be added to make it easy for developer to access them.

### DIFF
--- a/src/proto/constants.proto
+++ b/src/proto/constants.proto
@@ -1,0 +1,5 @@
+syntax = "proto3";
+
+enum CONST {
+  FARCASTER_EPOCH = 1609459200;
+}


vrypan:
Defining these values at the protobuf level allows easier upgrades to the protocol, if developers use them. For example, changing the size of casts, the max number of embeds, etc, will be automatic when developers upgrade to the newer version, if instead of defining their own constants (up to two embeds per cast, for example), they use the protocol constants.

sanjayprabhu:
Hmm, I like the idea, but not sure protobuf enums are the right fit. Maybe it can be a config file. Let me think about this.

vrypan:
Since we define the protocol in protobufs, this is the protobuff-native hack :-) to define constants... A config file will have to be fetched and parsed, etc, ie one more dependency and possible point of failure.

Issue #544, opened 2025-06-16
feat: Improve snapchain event visibility


sanjayprabhu:
Make it easier for snapchain clients to track how well they are keeping up with events and detect when they have missed events. To do this, we need the following:

- Add a `BLOCK_CONFIRMED` event type that's the first event emitted for a block. It should include the block number, shard id, timestamp, block hash and total number of events in the block
- Update GetEvents (and the http endpoint) to support start and end timestamps and block numbers

Related, but optional, it would be nice to have a way to determine which shard an fid belongs to so we can query for events on the correct shard. Maybe update `GetCurrentStorageLimitsByFid` to return the shard id or add a few `GetFid` endpoint.

PR #547, opened 2025-06-19
ABD-1 : Improve visibility into backend dropping snapchain events

abdulazeem-tk4vr:
https://github.com/farcasterxyz/snapchain/issues/544
Adds a "block confirmed" event to signal block finalization.
This event includes block metadata such as number, hash, timestamp, and total event count.
This commit also enables filtering events by type, timestamp, and block number in the HTTP API.

This provides a more reliable and informative event stream for clients and identify if any events were dropped/missed on the app layer side

### DIFF
--- a/src/storage/store/account/event.rs
+++ b/src/storage/store/account/event.rs
@@ -254,6 +254,10 @@ mod tests {
     fn test_block_confirmed_sequence_assignment() {
         let mut generator = HubEventIdGenerator::new(Some(123), Some(0));
 
+        // Test initial state
+        assert_eq!(generator.current_height, 123);
+        assert_eq!(generator.current_seq, 0);
+
         // Test BLOCK_CONFIRMED gets sequence 0
         let block_confirmed = create_test_event(HubEventType::BlockConfirmed);
         let event_id = generator.generate_id(&block_confirmed).unwrap();
@@ -261,6 +265,7 @@ mod tests {
 
         assert_eq!(height, 123);
         assert_eq!(sequence, 0);
+        assert_eq!(generator.current_seq, 0); // Should remain 0 for BLOCK_CONFIRMED
 
         // Test regular event gets sequence 1 (sequence 0 is skipped)
         let regular_event = create_test_event(HubEventType::MergeMessage);
@@ -269,6 +274,7 @@ mod tests {
 
         assert_eq!(height, 123);
         assert_eq!(sequence, 1);
+        assert_eq!(generator.current_seq, 2); // Should be incremented
 
         // Test another regular event gets sequence 2
         let regular_event2 = create_test_event(HubEventType::PruneMessage);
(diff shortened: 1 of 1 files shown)


src/network/server.rs
```diff
@@ -239,7 +239,7 @@ impl MyHubService {
         }
 
         let (tx, rx) = oneshot::channel();
-
+        println!("The incoming message is: {:?}", message);
```
aditiharini, line 242: Remove this

src/network/rpc_extensions.rs
```diff
@@ -119,6 +120,36 @@ impl LinksByFidRequest {
     }
 }
 
+impl EventsRequest {
+    pub fn timestamps(&self) -> (Option<u64>, Option<u64>) {
+        let start_timestamp = match self.start_timestamp {
```
aditiharini, line 125: Can you use `clone()`? If so, maybe not worth this level of indirection-- it's easy enough to call clone where needed.

src/network/server.rs
```diff
@@ -1008,8 +1009,67 @@ impl HubService for MyHubService {
                 events
             })
             .collect();
-        let combined_events: Vec<HubEvent> =
+        let mut combined_events: Vec<HubEvent> =
```
aditiharini, line 1012: Can you combine all the filtering into one pass? It's more efficient and I don't think it'll be messier.

aditiharini, line 1012: One thing to note here is that you're doing the filtering after you pull the events out of the db. This means that the user might receive empty responses with a pointer to a next page token to look at because there are no valid events in this page but there are still more events to look at. I think we should do this filtering while pulling events out of the db so we always return as full of a response as we can. Here's an example from another store: https://github.com/farcasterxyz/snapchain/blob/81c7e8ecbb6f110cacdf72a8d52d4bae3144dc9d/src/storage/store/account/onchain_event_store.rs#L301

aditiharini: There might be some complications with this because we don't store event timestamp and block number directly in the db. If that's a blocker, we should just start storing those values in the db.

src/storage/store/account/event.rs
```diff
@@ -211,3 +232,455 @@ impl HubEvent {
...
+
+        let (extracted_height, extracted_sequence) =
+            HubEventIdGenerator::extract_height_and_seq(event_id);
+        assert_eq!(extracted_height, height);
+        assert_eq!(extracted_sequence, sequence);
+
+        // Test with sequence that needs masking
+        let height = 999;
+        let sequence = 2u64.pow(SEQUENCE_BITS) + 100; // Should be masked
+        let event_id = HubEventIdGenerator::make_event_id(height, sequence);
```
aditiharini, line 366: Yeah, again this makes me question whether `make_event_id` should be public. In the `generate_id` function we have guardrails so that we error on sequence numbers which are too high.

src/storage/store/account/event.rs
```diff
@@ -211,3 +232,455 @@ impl HubEvent {
...
+        let result = generator.generate_id(&regular_event);
+
+        assert!(result.is_err());
+        let error = result.unwrap_err();
+        assert!(error.message.contains("sequence cannot fit in event id"));
+        assert!(error.message.contains(&SEQUENCE_BITS.to_string()));
+    }
+
+    #[test]
+    fn test_store_event_handler_concurrent_access() {
```
aditiharini, line 558: Are there places where we have concurrent transactions in this change?

src/storage/store/engine.rs
```diff
@@ -1472,10 +1520,33 @@ impl ShardEngine {
...
+                // BLOCK_CONFIRMED event to the existing events before committing.
+                // This ensures every block gets a BLOCK_CONFIRMED event regardless of
+                // whether it goes through the full replay or uses cached results.
+                match self.add_block_confirmed_event(
+                    shard_chunk,
+                    cached_txn.events,
+                    &mut cached_txn.txn,
+                ) {
+                    Ok(events_with_block_confirmed) => {
+                        self.commit_and_emit_events(
```
aditiharini, line 1537: Can you move `add_block_confirmed_event` to commit_and_emit_events given it's called before that in both places?

src/storage/store/engine.rs
```diff
@@ -1856,3 +1955,220 @@ impl BlockEngine {
         }
     }
 }
+
+#[cfg(test)]
+mod tests {
```
aditiharini, line 1960: Should move to `engine_tests`

src/storage/store/engine_tests.rs
```diff
@@ -1246,9 +1263,12 @@ mod tests {
         test_helper::assert_contains_all_messages(messages, &[cast3]);
 
         // We receive a merge event for the add and the intermediate remove, even though it would never get committed to the db
-        assert_merge_event(&event_rx.try_recv().unwrap(), &cast1, 0);
-        assert_merge_event(&event_rx.try_recv().unwrap(), &cast2, 1);
-        assert_merge_event(&event_rx.try_recv().unwrap(), &cast3, 2);
+        // First receive BLOCK_CONFIRMED event
+        let _block_confirmed_event = &event_rx.try_recv().unwrap();
```
aditiharini, line 1267: I think it makes sense to add a `recv_next_event` function which takes in an param to skip block confirmed events and events rx and skips the block confirmed events if indicated.

src/network/server.rs
aditiharini: Yeah, I think we should get this change in along with this feature. I don't necessarily see this as an optimization-- more of a UX improvement for users calling the rpc. They may have to make a lot of rpc calls to get a very small amount of data unless we try to fill pages.

PR #555, opened 2025-06-24
feat: add event counts per event type to BlockConfirmed

aditiharini:
This is required to help the client determine if there are missed events with filtering applied on the event stream.


src/proto/hub_event.proto
```diff
@@ -46,6 +46,7 @@ message BlockConfirmedBody {
   uint64 timestamp = 3;
   bytes block_hash = 4;
   uint64 total_events = 5;
+  map<int32, uint64> total_events_by_type = 6;
```
sanjayprabhu, line 49: `event_counts_by_type`?

sanjayprabhu (approved):

Issue #476, opened 2025-05-13
Message timestamp specification is not enforced


BlinkyStitt:
The spec says that messages cannot be outside of 600 seconds from the current time. This was enforced by hubble.

But snapchain allowed messages with a timestamp far in the future.

https://explorer.neynar.com/0xc16f624d58459cc6b8d0d05512b3290b07571933

This has a timestamp of 2076. not sure if they did that on purpose or not, but the network should have rejected it.

multiple casts all from the same user:
0xc16f624d58459cc6b8d0d05512b3290b07571933
0x41f5295a3e81201654a64344a5a73df7eff5e349
0x80253aaca798ac1c0eec52b9afbf6217f8a54712
0x8b52dec545841f0e3f4fb42a9717e5b0d24ef10c
0x457c2bb5b90dec11e378a9da352954794bf4c5b4
0xfd70aaf6d11dcccb86a11cfd645586997a380f32

Alternatively, the spec is out of date. https://github.com/farcasterxyz/protocol/blob/main/docs/SPECIFICATION.md#2-message-specifications

> m.data.timestamp must be a valid Farcaster epoch timestamp not more than 600 seconds ahead of the current time.

sanjayprabhu:
This was required for the migration, we'll start enforcing this soon now that hubs are out of the picture.

PR #569, opened 2025-07-08
fix: check for timestamp being too far in the future

aditiharini:
Closes https://github.com/farcasterxyz/snapchain/issues/476

### DIFF
--- a/src/core/validations/message.rs
+++ b/src/core/validations/message.rs
@@ -1,3 +1,4 @@
+use crate::core::util::get_farcaster_time;
 use crate::core::validations::error::ValidationError;
 use crate::core::validations::validate_cast_id;
 use crate::proto::{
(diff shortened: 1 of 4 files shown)


src/core/validations/message.rs
```diff
@@ -64,6 +67,22 @@ fn validate_frame_action_body(body: &FrameActionBody) -> Result<(), ValidationEr
     Ok(())
 }
 
+pub fn validate_timestamp(
+    message_timestamp: u32,
+    engine_version: EngineVersion,
+) -> Result<(), ValidationError> {
+    if engine_version.is_enabled(ProtocolFeature::FutureTimestampValidation) {
+        let current_time = get_farcaster_time()? as u32;
```
sanjayprabhu, line 75: Should this use the block timestamp?

aditiharini: Yeah actually I think that's better so that the check is deterministic.

sanjayprabhu (approved):

PR #570, opened 2025-07-08
feat: Storage redenomination

sanjayprabhu:
Implements the Storage Re-denomination FIP: https://github.com/farcasterxyz/protocol/discussions/229

As part of it, we're also extending the expiry of storage units that are set to expire on Aug 29 2025 by one year.


sanjayprabhu:
We'll keep it as is for now, because reducing it too much below 100 could cause new users signing up to hit limits. Setting up your profile, following a few people, a cast and a few reactions/replies could easily cross 25-30 messages. We should monitor first to see if it's problem.

PR #580, opened 2025-07-14
automated attempt to find docs inconsistencies

deodad:


site/docs/pages/guides/migrating-to-snapchain.mdx
```diff
@@ -19,7 +19,7 @@ Snapchain is designed to be a drop-in replacement for [Hubble](https://github.co
 
 Messages once submitted must be included in blocks, similar to blockchain transactions. The `submitMessage` has two main differences from Hubble: 
 
-1. `submitMessage` requests must all contain `dataBytes` for Snapchain. The `hub-nodejs` builders handle this in all versions `>=0.13.0`, but if you're not using those you will need to manually update this like so:
+1. `submitMessage` requests can optionally include `dataBytes` for better performance. If not provided, Snapchain will automatically encode the `data` field. The `hub-nodejs` builders handle this optimization in all versions `>=0.13.0`. If you want to manually optimize this:
```
aditiharini, line 22: We intentionally recommend that all clients provide dataBytes. If the client uses a different proto serialization impl than the server, we run into invalid hash in the server because client and server compute the hash off different bytes. To avoid such issues, all clients should provide dataBytes.

site/docs/pages/guides/writing-messages.mdx
```diff
@@ -85,6 +113,25 @@ and return early if so.
 If you don't have a funded account you can use, note the address and private key pair that's logged. Transfer funds to the address and use the same private key as the `CUSTODY_PRIVATE_KEY` for the next run of the script. 
 
 ```typescript
+// Create wallet and public clients
+const account = privateKeyToAccount(CUSTODY_PRIVATE_KEY as `0x${string}`);
+const accountPrivateKey = CUSTODY_PRIVATE_KEY === '<REQUIRED>' ? generatePrivateKey() : CUSTODY_PRIVATE_KEY;
```
aditiharini, line 118: This update seems weird? accountPrivateKey should be defined before account and then we should use accountPrivateKey instead of CUSTODY_PRIVATE_KEY to initialize account.

site/docs/pages/reference/datatypes/events.md
```diff
@@ -8,11 +8,14 @@ Snapchain keeps events around for 3 days after which they are deleted to save sp
...
+| shard_index  | [uint32](#uint32)                                                                                                                                                                                                                                                                                                                          |       | The shard index where the event occurred                          |
+| timestamp    | [uint64](#uint64)                                                                                                                                                                                                                                                                                                                          |       | The timestamp when the event was created                          |
```
aditiharini, line 17: I would describe this as "The timestamp when the block that produced this event was created". The current description is a bit ambiguous/misleading.

PR #587, opened 2025-07-16
feat: expose next engine version timestamp in GetInfo API

### LINKED ISSUE #586: feat: Return upcoming engine version schedule in GetInfo
Via @christopherwxyz

Would be nice to be able to query the upcoming protocol schedule from within snapchain. Node operators can query validators or other trusted nodes to know if they are behind. We do return the `version` string already, but having the protocol upgrade schedule would be nice to have as well.

christopherwxyz:
- Add next_engine_version_timestamp field to GetInfoResponse proto
- Implement next_version_timestamp_for() method in EngineVersion
- Update gRPC GetInfo handler to include next engine version timestamp
- Update HTTP API InfoResponse struct and mapping function
- Add comprehensive tests for the new functionality

Resolves #586

### DIFF
--- a/src/network/http_server.rs
+++ b/src/network/http_server.rs
@@ -316,6 +316,8 @@ pub struct InfoResponse {
     pub version: String,
     #[serde(rename = "peer_id")]
     pub peer_id: String,
+    #[serde(rename = "nextEngineVersionTimestamp")]
+    pub next_engine_version_timestamp: u64,
 }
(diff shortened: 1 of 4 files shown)


sanjayprabhu:
Thanks, rebased and fixed tests here https://github.com/farcasterxyz/snapchain/pull/595. Will merge that one.

Issue #510, opened 2025-05-29
feat: Design and implement a new sync model that doesn't rely on full block history


sanjayprabhu:
Snapchain currently syncs in two ways:
 - Snapshot sync which syncs block state and account state at the same time by downloading a copy of the db
 - block sync which syncs and replays blocks to regenerate current account state from historical transactions

Because snapchain design intends to throw away vast majority of historical blocks, we need an additional sync mechansim that's decentralized and works with nodes that don't have full block history.

Ideally, it would work like this:
1. When you start a brand new node, it fetches the latest block.
2. For all accounts in that block, atomically fetch the account state as of that block via a new API
3. Persist account state, verify the merkle roots match for the account and proceed to the next block and repeat the process.
4. Deal with complications like rapidly changing accounts, byzantine peers, etc

aditiharini:
@topocount this needs to be rolled out.

PR #589, opened 2025-07-18
Read implementation for replication

defnotpaul:
Towards: #510

Adds:

- Extension to Engine to support replaying individual transactions for backfilling FID state data
- ReplicationStores: used to hold "snapshots" (read-only views) into individual shard stores taken at a specified block interval. These snapshots are utilized to provide a consistent view into FID data at known block height(s).
- Replicator: the business logic for taking snapshots and exporting FID data
- Replication Service: a new grpc service that exposes an interface in front of replicator for querying FID state.

Follow-ups:

- [ ] Some accounts have a lot of messages (e.g. fids 2 and 3) so we'll need to sync them over multiple requests.
- [ ] Solve for message ordering at the engine level [:link:](https://github.com/farcasterxyz/snapchain/pull/589#discussion_r2216488125)
- [ ] Allow db-free lookups at the engine level (e.g. when an FID doesn't exist we shouldn't error)
- [ ] Don't commit transactions for FID unless the account root matches [:link:](https://github.com/farcasterxyz/snapchain/pull/589#discussion_r2221348609)


src/storage/db/rocksdb.rs
```diff
@@ -290,34 +193,44 @@ impl RocksDB {
         Ok(())
     }
 
-    pub fn db(&self) -> RwLockReadGuard<'_, Option<TransactionDB>> {
-        self.db.read().unwrap()
+    pub fn db(&self) -> RwLockReadGuard<'_, Option<DBProvider>> {
```
defnotpaul, line 196: We should verify that callers of this **only** use it for read-only.

src/storage/store/replication/replicator.rs
```diff
@@ -0,0 +1,306 @@
...
+        if block_number % self.snapshot_options.interval != 0 {
+            return Ok(());
+        }
+
+        // Open a snapshot
+        self.stores.open_snapshot(block_number, msg.shard_id)
+    }
+}
+
+fn collect_messages_with_page_options<
```
aditiharini, line 116: Some accounts have a lot of messages (e.g. fids 2 and 3) so we'll need to sync them over multiple requests.

src/storage/store/replication/replicator.rs
```diff
@@ -0,0 +1,306 @@
...
+    })
+}
+
+fn collect_compact_state<T: StoreDef + Clone>(store: &Store<T>, fid: &u64) -> Vec<proto::Message> {
+    collect_messages_with_page_options(|page_options| {
+        store.get_compact_state_messages_by_fid(*fid, page_options)
+    })
+}
+
+fn build_user_messages_for_fid(stores: &Stores, fid: u64) -> Vec<proto::Message> {
```
aditiharini, line 155: How are we handling the ordering problem here? Verifications should be committed first, then username proofs, then everything else right?

src/storage/store/replication/post_commit_handler.rs
```diff
@@ -0,0 +1,59 @@
...
+        }
+    }
+}
+
+pub struct PostCommitHandler {
+    pub sender: tokio::sync::mpsc::Sender<PostCommitMessage>,
+    handle: tokio::task::JoinHandle<()>,
+}
+
+impl PostCommitHandler {
```
aditiharini, line 40: Is this used anywhere right now? I'm having a little trouble following where this fits in. I thought this recv would be happening closer to where the snapshot is taken.

src/storage/store/replication/replicator.rs
defnotpaul: Good call. We've chatted about this, and we're deferring on implementing handling for this PR.

src/storage/store/replication/replicator.rs
```diff
@@ -0,0 +1,374 @@
...
+            }
+        };
+
+        // Clean up old snapshots
+        let expiration_block = block_number.saturating_sub(self.snapshot_options.max_age);
+        self.stores.close_snapshots_below(expiration_block);
+
+        // Check if we can take a snapshot
+        if block_number % self.snapshot_options.interval != 0 {
+            return Ok(());
```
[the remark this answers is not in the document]
defnotpaul: We've addressed this by utilizing time-based-expiry for snapshots

src/storage/store/replication/replicator.rs
```diff
@@ -0,0 +1,374 @@
...
+                user.push(user_tx);
+            }
+        }
+
+        Ok((sys, user))
+    }
+
+    pub fn handle_post_commit_message(
+        &self,
+        msg: &PostCommitMessage,
```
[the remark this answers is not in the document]
defnotpaul: The callback is in the `run` function. We're intentionally trying to keep the messaging logic separate from the business logic here, so that's handled elsewhere.

src/storage/store/replication/replicator.rs
```diff
@@ -0,0 +1,374 @@
...
+pub struct ReplicatorSnapshotOptions {
+    pub interval: u64, // Interval in blocks to take snapshots
+    pub max_age: u64,  // Maximum age of snapshots in blocks
+}
+
+impl Default for ReplicatorSnapshotOptions {
+    fn default() -> Self {
+        ReplicatorSnapshotOptions {
+            interval: 1_000, // Default to taking a snapshot every 1000 blocks
+            max_age: 10_000, // Default to keeping snapshots for 10000 blocks
```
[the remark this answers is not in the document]
defnotpaul: Updated these -- note: the real values that are utilized reside in `src/cfg.rs` now.

src/storage/store/engine.rs
```diff
@@ -1504,33 +1515,58 @@ impl ShardEngine {
...
+                    error!("Post commit hook send failed: {}", err);
+                    return;
+                }
+
+                // Attempt to wait for the receiver to process the message, but don't block
+                // indefinitely in case the receiver is not ready or available.
+                match tokio::time::timeout(Duration::from_millis(200), oneshot_rx).await {
+                    Ok(Ok(_)) => {} // success
+                    Ok(Err(err)) => error!("Post commit hook failed: {}", err),
+                    Err(err) => error!("Post commit hook receive failed: {}", err),
```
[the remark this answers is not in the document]
defnotpaul: Ya, we're optimistically waiting for a response to come back. If not, we proceed, but we're doing our best to ensure the snapshot can be opened without writes coming in.

src/storage/store/engine.rs
```diff
@@ -1551,6 +1587,17 @@ impl ShardEngine {
...
+
+            let count_fn = Self::make_count_fn(self.statsd_client.clone(), self.shard_id);
+            let count_callback = move |read_count: (u64, u64)| {
+                count_fn("trie.db_get_count.total", read_count.0);
+                count_fn("trie.db_get_count.for_commit", read_count.0);
+                count_fn("trie.mem_get_count.total", read_count.1);
+                count_fn("trie.mem_get_count.for_commit", read_count.1);
+            };
+            let trie_ctx = &merkle_trie::Context::with_callback(count_callback);
+            // let trie_ctx = &merkle_trie::Context::new();
```
defnotpaul, line 1599: delete me

src/storage/store/engine.rs
```diff
@@ -1504,33 +1515,58 @@ impl ShardEngine {
...
+    async fn post_commit(&mut self, header: proto::ShardHeader) {
+        match &mut self.post_commit_tx {
+            None => return,
+            Some(tx) => {
+                // Attempt to send the post commit message and wait for a response,
+                // but don't block indefinitely in case the receiver is not ready, unavailable,
+                // or too slow.
+
+                let (oneshot_tx, oneshot_rx) = oneshot::channel();
+                let send_future = tx.send(PostCommitMessage {
```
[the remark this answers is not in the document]
defnotpaul: See comment below

src/storage/store/replication/replicator.rs
defnotpaul: Added to our follow-ups list

sanjayprabhu (commented):
Looks pretty good, just a few minor comments. 

Have you tried running a devnet network locally with and without snapshotting?

src/main.rs
```diff
@@ -453,6 +524,14 @@ async fn main() -> Result<(), Box<dyn Error>> {
     } else {
         let (shard_decision_tx, shard_decision_rx) = broadcast::channel(100);
 
+        // Setup post-commit channel if replication is enabled
+        let (engine_post_commit_tx, engine_post_commit_rx) = if app_config.replication.enable {
+            let (tx, rx) = mpsc::channel::<PostCommitMessage>(1);
```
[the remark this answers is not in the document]
defnotpaul: Ya, left a TODO asking if making this buffer larger (e.g. number of shards) would be better approach.

sanjayprabhu (approved):
Nice, looks good.

PR #593, opened 2025-07-24
chore: add more metrics for events emitted

aditiharini:
Add metrics so we can determine how often prunes and revokes are happening and how often messages of various types are being merged.


src/storage/store/engine.rs
```diff
@@ -1526,6 +1518,28 @@ impl ShardEngine {
         );
         self.gauge("max_messages_per_block", self.max_messages_per_block as u64);
 
+        for event in events {
+            self.count(
+                "commit.emitted_event",
```
sanjayprabhu, line 1523: Nice, this will be useful.

sanjayprabhu (approved):

PR #598, opened 2025-07-27
chore: add tests to ensure that snapchain validations are in sync with client validations

aditiharini:
Add tests that run the same messages through both the snapchain validations and the client side validations. This will help us keep validations in sync between snapchain and the client library without having to use the rust code directly in the client library.


src/core/validations/client_parity_tests/generate-messages.ts
```diff
@@ -0,0 +1,219 @@
+#!/usr/bin/env ts-node
```
sanjayprabhu, line 1: We should put this into a separate directory that's outside of `src` so we don't mix rust and typescript code. Maybe `scripts/validations`?

src/core/validations/client_parity_tests/generate-messages.ts
```diff
@@ -0,0 +1,219 @@
...
+
+  const testInputsJson = testInputs.map((testInput) => {
+    return {
+      message_file: `${outputDir}/${testInput.filename}`,
+      is_valid: testInput.isValid
+    }
+  });
+
+  fs.writeFileSync(
+    path.join(__dirname, 'test_inputs.json'),
```
sanjayprabhu, line 208: Instead of separate files, can we encode to hex and keep it in the same file as the json?  Might be easier to manage. Don't feel too strongly though

sanjayprabhu (commented):
Nice, this is great.

PR #599, opened 2025-07-28
fix: Add progress info to snapshot download

### LINKED ISSUE #380: feat: Show snapshot progress when downloading
We should show a progress bar of some kind, and ideally and ETA when downloading snapshots

adityapk00:
Fixes #380 

Add progress info when downloading the initial snapshot. If we're running in interactive mode, it shows:
```
⠉ Overall Progress [00:00:24] [>---------------------------------------] 2/3572 chunks (11h)
⠲ chunk_0003.bin                 [====>-----------------------------------] 9.67 MiB/86.90 MiB (2.80 MiB/s)
```

If running in non-interactive mode (eg in docker-compose), it shows
```
snapshot: Completed download of chunk_0002.bin (87310498 bytes). Progress: 2/3572 chunks (0%) (ETA: 5h57m)
snapshot: Downloading zipped snapshot chunk chunk_0003.bin for shard 1 (3/1766 chunks in shard) Shard ETA: 2h56m
```

### DIFF
--- a/src/storage/db/snapshot.rs
+++ b/src/storage/db/snapshot.rs
@@ -286,6 +286,7 @@ pub async fn download_snapshots(
 
     // Check if we're in a TTY environment - if not, disable progress bars
     let use_progress_bars = atty::is(atty::Stream::Stdout);
+    let global_start = SystemTime::now();
     if !use_progress_bars {
         info!("No TTY detected, progress will be logged instead of shown as progress bars");
         info!(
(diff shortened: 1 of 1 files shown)


sanjayprabhu (approved):
Nice. There's a lot of improvements we can make here (parallel downloads, resumable downloads, improved retries) but snapdown already has these features, so I think it's not worth us spending too much effort here, and we should just use the tool the community build. 

Longer term, once we have the new sync feature https://github.com/farcasterxyz/snapchain/issues/510, we might even be able to get rid of snapshots. 

Feel free to merge or close.

PR #600, opened 2025-07-28
feat: Add SubmitBulkMessages rpc/http API

adityapk00:
Add the SubmitBulkMessages API (which had been commented out). 

The endpoint (both grpc and http API) accepts an array of messages, and submits them sequentially. Right now, the messages need to be independent (i.e., a message that depends on a previous message in the array to succeed is not supported yet)

It will attempt to merge all the messages, and return success/failure for each one individually

### DIFF
--- a/site/docs/pages/reference/httpapi/message.md
+++ b/site/docs/pages/reference/httpapi/message.md
@@ -54,10 +54,62 @@ curl -X POST "http://127.0.0.1:3381/v1/submitMessage" \
 }
 ```
 
+
+## submitBulkMessages
+
+Submit several signed protobuf-serialized messages to the Hub at once. Each one will be submitted to the node sequentially.
+
+**Query Parameters**
+| Parameter | Description                         | Example |
+| --------- | ----------------------------------- | ------- |
+|           | This endpoint accepts no parameters |         |
+
+**Example**
+
+```bash
+curl -X POST "http://127.0.0.1:3381/v1/submitBulkMessages" \
+     -H "Content-Type: application/octet-stream" \
+     --data-binary "@SubmitBulkMessagesRequest.encoded.protobuf"
+
+```
+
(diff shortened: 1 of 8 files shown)


src/network/server.rs
```diff
@@ -591,6 +591,51 @@ impl HubService for MyHubService {
...
+
+        let mut results = Vec::with_capacity(num_messages);
+        for mut msg in messages {
+            // Ensure data_bytes are decoded before processing
+            message_bytes_decode(&mut msg);
+            let message_hash_for_error = msg.hash.clone();
+
+            // The `submit_message_internal` is async, so we await it so messages are processed sequentially.
+            // We don't want to do parallel processing here, as some message may depend on a previous message.
+            let result = self.submit_message_internal(msg, false).await;
```
sanjayprabhu, line 615: Note that this doesn't actually wait for the messages to be included in the block, it only simulates it. So if messages depend on each other, they will be simulated independently.

Which will probably become an issue when we get to the next step and support merging both messages at once, we want the simulation to include both messages at the same time. So I think we need to change `simulate_message` to take an array and pass it all the way through.

Also, we should remove the `bypass_validation` flag, it was only used for early performance testing and no one is passing in true anymore.

site/docs/pages/reference/httpapi/message.md
```diff
@@ -54,10 +54,62 @@ curl -X POST "http://127.0.0.1:3381/v1/submitMessage" \
 }
 ```
 
+
+## submitBulkMessages
+
+Submit several signed protobuf-serialized messages to the Hub at once. Each one will be submitted to the node sequentially.
```
sanjayprabhu, line 60: Nice. We've been talking about getting claude to do this automatically for us.

sanjayprabhu (approved):

PR #602, opened 2025-07-29
feat: Allow multi message simulation with dependent messages

adityapk00:
Allow dependent messages during bulk submit. 

Dependent messages are when one message in the array of messages being submitted depends on an earlier message in the same array. For eg., submitting a `[ENS proof , UserDataAdd]` message[] , which allows both messages to be accepted and submitted in one go.  See the `test_simulate_bulk_messages_username_proof_and_set` for an example 

The main change is here is to pass the DB txn all the way down the stores, so that they can use the current transaction + DB (instead of just the DB) to validate/process messages. 

This PR allows ENS proofs to be submitted along with their dependent messages. (Other types of dependent messages (onchain events, verifications, fnames etc....) are in a separate PR

### DIFF
--- a/src/storage/store/engine.rs
+++ b/src/storage/store/engine.rs
@@ -1287,6 +1287,7 @@ impl ShardEngine {
             let proof_message = UsernameProofStore::get_username_proof(
                 &self.stores.username_proof_store,
                 &name.as_bytes().to_vec(),
+                txn,
             )
             .map_err(|e| MessageValidationError::StoreError(e))?;
             match proof_message {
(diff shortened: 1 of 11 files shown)


src/storage/store/engine.rs
```diff
@@ -1707,6 +1709,86 @@ impl ShardEngine {
...
+                .entry(fid)
+                .or_insert_with(Vec::new)
+                .push((index, message));
+        }
+
+        // Initialize results vector with placeholder values
+        let mut results: Vec<Result<(), MessageValidationError>> = vec![Ok(()); messages.len()];
+
+        // Process each FID group
+        for (fid, fid_messages) in messages_by_fid {
```
aditiharini, line 1737: What's the benefit of doing this per-fid rather than doing all the messages in a single transaction?

PR #608, opened 2025-07-31
feat: Add support for onchain events and verification messages in a single transaction

adityapk00:
This PR builds on top of #602 with additional support for on chain messages and user verifications. 

With this, we can now submit [Register Id, Add Storage, add Signer, add Cast] in the same transaction, and they will al get accepted. A block proposer can also inclulde all of these in the same block. See `engine_tests.rs:test_bulk_register_add_signer_and_cast_commit` for an example of how this works.


src/storage/store/engine.rs
```diff
@@ -1243,31 +1318,62 @@ impl ShardEngine {
...
+        }
 
-        // Check that signer is valid
-        self.stores
-            .onchain_event_store
-            .get_active_signer(message_data.fid, message.signer.clone())
-            .map_err(|_| MessageValidationError::MissingSigner)?
-            .ok_or(MessageValidationError::MissingSigner)?;
+        // 2. Check that the user has a valid signer
+        match txn_onchain_state.get_effective_signer_state(message_data.fid, &message.signer) {
```
sanjayprabhu, line 1335: Why do this vs passing the txn batch to `get_active_signer`? I think this should be the same, but I'm not sure if there are some edges around event ordering where this map may not match what's actually in the db.

sanjayprabhu (commented):
Looks good to me. One question though. 

Also, we're going to do a release tomorrow, so let's merge this in after the release.

src/storage/store/engine.rs
adityapk00: So i put it here to support this (admittedly contrived) scenario - If there is a SignerRemove message, and then a CastAdd that uses the signer in the same batch, then the CastAdd should fail. The `get_effective_signer_state` handles the signer removes, but it would be weird to have that be in get_signer_add. 

Don't have a strong opinion here, so if you think this is too contrived, I can switch it into get_signer_add

aditiharini: I have the same question as Sanjay here. I thought this would be implemented by threading the currently active txn through to more stores. 

I'm not sure I understand what the issue is with this approach even in the scenario you described. `get_active_signer` should account for signer removes too.

PR #614, opened 2025-08-04
chore: add logs for onchain events subscription exiting

aditiharini:
Add logs to understand if the onchain events subscription is exiting.


sanjayprabhu (approved):
May also be worth adding a metric inside `stream.next()` within `sync_live_events`

aditiharini:
replaced

PR #619, opened 2025-08-06
feat: decouple shard 0 block production from other shards block production

aditiharini:
The approach is
- If the shard witness for the next height after the last recorded shard witness is available, use it. 
- If not, carry over the previous shard witness. 

This makes it easy to figure out what the last recorded shard witness for each shard is (you can look at the previous block) and doesn't introduce any complexity around validating multiple shard chunks for a single block. It also maintains the invariant that the number of shard witnesses in each block matches the number of shards. 

The timeout we spend waiting for the shard witness has been removed. We don't need this because we wait a whole block time anyway between blocks and we have a timeout for the propose step individually too. Once shard 0 starts containing data, it's annoying to have this timeout because it eats into the time available to actually process the data.


src/consensus/proposer.rs
```diff
@@ -350,59 +351,100 @@ impl BlockProposer {
...
+            for (shard_id, store) in self.shard_stores.iter() {
+                let last_shard_witness = self.engine.get_last_shard_witness(height, *shard_id);
+
+                // This is the height we want the witness for
+                let desired_shard_block_number = if height.block_number == 1 {
+                    1
+                } else if let Some(ref last_shard_witness) = last_shard_witness {
+                    last_shard_witness.height.unwrap().block_number + 1
+                } else {
+                    // Error case, we should never get here
```
sanjayprabhu, line 394: We should log here

src/consensus/malachite/spawn.rs
```diff
@@ -202,6 +202,7 @@ impl MalachiteConsensusActors {
         .await?;
         let sync_config = ValueSyncConfig {
             request_timeout: config.sync_request_timeout,
+            status_update_interval: config.sync_status_update_interval,
```
sanjayprabhu, line 205: Why do we need this? Is it related to this change?

src/consensus/proposer.rs
```diff
@@ -350,59 +351,100 @@ impl BlockProposer {
...
+                    continue;
+                };
+
+                if let Ok(Some(next_shard_witness)) =
+                    self.get_shard_witness_at_height(desired_shard_block_number, store)
+                {
+                    chunks.insert(*shard_id, next_shard_witness);
+                } else if let Some(last_shard_witness) = last_shard_witness {
+                    // If the next shard witness is not available, carry over the previous one
+                    chunks.insert(*shard_id, last_shard_witness);
```
sanjayprabhu, line 404: We should emit a metric or log line here so we know if this happens.

sanjayprabhu (commented):
I wonder if we should run shard 0 at a slightly faster block interval so it can catch up to the other shards now. As it is, if shard 0 ever falls behind, it will never catch up.

src/consensus/malachite/spawn.rs
aditiharini: No it's not related to this change. It's to speed up the tests. The tests that were exercising sync were taking a really long time to run. I think it's because peers weren't sending status updates often enough. This seemed to cut that down (though some of these tests are still really slow for reasons I haven't been able to figure out yet).

sanjayprabhu (dismissed):
LGTM. Let's ticket speeding up Shard 0 block production. That can happen once this goes live, doesn't require a protocol change.

PR #629, opened 2025-08-14
chore: extract mempool polling logic into a module for sharing

aditiharini:
We will use this same logic to poll from the mempool for shard 0.

### DIFF
--- a/src/storage/store/mempool_poller.rs
+++ b/src/storage/store/mempool_poller.rs
@@ -1,0 +1,149 @@
+use crate::mempool::mempool::MempoolMessagesRequest;
+use crate::proto::{self, FarcasterNetwork, Transaction};
+use crate::storage::store::account::OnchainEventStorageError;
+use crate::utils::statsd_wrapper::StatsdClientWrapper;
+use itertools::Itertools;
+use std::string::ToString;
+use std::time::Duration;
+use thiserror::Error;
+use tokio::sync::{mpsc, oneshot};
+use tokio::time::timeout;
+use tracing::{error, info};
+
+#[derive(Error, Debug)]
+pub enum MempoolPollerError {
+    #[error(transparent)]
+    MessageReceiveError(#[from] oneshot::error::RecvError),
+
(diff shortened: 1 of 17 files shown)


src/storage/store/mempool_poller.rs
```diff
@@ -0,0 +1,149 @@
...
+        };
+        proto::MempoolMessage {
+            mempool_message: Some(msg),
+        }
+    }
+}
+
+impl MempoolPoller {
+    pub fn for_read_node(&self) -> bool {
+        self.messages_request_tx.is_none()
```
sanjayprabhu, line 59: This has become a very roundabout way for the engine to determine if it's a read node. Might be fine for now, but wonder if we should just pass in a boolean to the engine instead.

src/storage/store/mempool_poller.rs
```diff
@@ -0,0 +1,149 @@
...
+    pub fn create_transactions_from_mempool(
+        &mut self,
+        messages: Vec<MempoolMessage>,
+    ) -> Result<Vec<Transaction>, MempoolPollerError> {
+        let mut transactions = vec![];
+
+        let grouped_messages = messages.iter().into_group_map_by(|msg| msg.fid());
+        let unique_fids = grouped_messages.keys().len();
+        for (fid, messages) in grouped_messages {
+            let mut transaction = Transaction {
```
sanjayprabhu, line 119: This living outside the engine and having the logic split between here and the engine feels weird. Again, might be fine, just wanted to make a node. I think the block engine might have different rules on which messages to include (e.g. no user messages except storage lend) so this split might be necessary, still feels a little weird.

sanjayprabhu (approved):

src/storage/store/mempool_poller.rs
aditiharini: I'm not really sure why we're handling storage being missing in this way. I think we should be running this check inside  replay_snapchain_transaction the same way we check for active signer there.

PR #634, opened 2025-08-15
feat: produce transactions and events in shard 0

aditiharini:


src/storage/store/block_engine.rs
```diff
@@ -0,0 +1,444 @@
...
+            let mut event = BlockEvent {
+                seqnum: event_seqnum,
+                hash: vec![],
+                r#type: BlockEventType::Heartbeat as i32,
+                block_number: height.block_number,
+                event_index: events.len() as u64,
+                block_timestamp: timestamp.to_u64(),
+                body: Some(block_event::Body::HeartbeatEventBody(HeartbeatEventBody {})),
+            };
+            event.hash = blake3::hash(event.encode_to_vec().as_slice())
```
aditiharini, line 127: Validate that this is the right way to generate a hash.

src/consensus/proposer.rs
```diff
@@ -489,6 +490,14 @@ impl Proposer for BlockProposer {
             .collect_confirmed_shard_witnesses(height, version, timeout)
             .await;
 
+        let mempool_timeout = Duration::from_millis(200);
+        let messages = self
```
aditiharini, line 494: Nothing in the mempool currently gets routed to shard 0. Maybe worth just defaulting to an empty list of transactions rather than having this in here. 

We should test that the engine code works with non-empty transactions though.

src/storage/store/block_engine.rs
```diff
@@ -0,0 +1,444 @@
...
+                .as_bytes()
+                .to_vec();
+            // Store these events so
+            // (1) It's possible to figuure out the max seqnum easily
+            // (2) It's possible to query over them in an rpc and see what has been produced.
```
aditiharini, line 132: Need to add this RPC

src/storage/store/block_engine.rs
```diff
@@ -0,0 +1,549 @@
...
+        shard_root: &[u8],
+        events_hash: &Vec<u8>,
+        source: ProposalSource,
+        height: Height,
+        timestamp: &FarcasterTime,
+    ) -> Result<(), BlockEngineError> {
+        let now = std::time::Instant::now();
+        // Validate that the trie is in a good place to start with
+        match self.get_last_block() {
+            None => { // There are places where it's hard to provide a parent hash-- e.g. tests so make this an option and skip validation if not present
```
sanjayprabhu, line 310: I'm realizing we should probably only do this if we're in a test env (or devnet). Something for later.

src/storage/store/block_engine.rs
```diff
@@ -0,0 +1,549 @@
...
+                // Store these events so
+                // (1) It's possible to figuure out the max seqnum easily
+                // (2) It's possible to query over them in an rpc and see what has been produced.
+                self.block_event_store.put_block_event(&event, txn).unwrap();
+                events.push(event);
+            }
+        }
+
+        let events_hash = if events.is_empty() {
+            vec![]
```
sanjayprabhu, line 181: Hmm, I'm not sure if empty event should hash to empty. What's the value returned by the hasher if you don't call any `update`s on it?

src/storage/store/account/block_event_store.rs
```diff
@@ -0,0 +1,158 @@
...
+    txn.put(primary_key.clone(), block_event.encode_to_vec());
+    Ok(())
+}
+
+#[derive(Default, Clone)]
+pub struct BlockEventStore {
+    pub db: Arc<RocksDB>,
+}
+
+impl BlockEventStore {
```
sanjayprabhu, line 135: We might need and API for this.

We should also prune these when the corresponding blocks are pruned. 

Worth ticketing both.

src/consensus/proposer.rs
```diff
@@ -492,8 +493,16 @@ impl Proposer for BlockProposer {
         round: Round,
         timeout: Duration,
     ) -> FullProposal {
-        let timestamp = FarcasterTime::current();
-        let version = EngineVersion::version_for(&timestamp, self.network);
+        let mempool_timeout = Duration::from_millis(200);
+        let messages = self
+            .engine
+            .mempool_poller
+            .pull_messages(mempool_timeout)
```
sanjayprabhu, line 500: We also need to duplicate onchain events to shard 0 in the mempool right? I don't think it has a concept of shard 0 right now?

sanjayprabhu (commented):
We might also want to add a consensus level test that checks that onchain events are present in block 0 and it's generating heartbeat events.

src/consensus/proposer.rs
aditiharini: Yeah, I didn't want to add that here because I don't want onchain events to get into shard 0 blocks yet. I wanted to test that the system works with just heartbeats first. We can route onchain events to shard 0 without an engine version upgrade later once we're confident the system works. The reason I added support for onchain events in the engine is so that I can test all the code that deals with the trie. I wanted to see that the shard root updated as expected. Added a comment to remember to route onchain events to shard 0.

src/storage/store/block_engine.rs
aditiharini: Added a comment

aditiharini:
> We might also want to add a consensus level test that checks that onchain events are present in block 0 and it's generating heartbeat events.

I ended up adding this consensus test in the next feature: https://github.com/farcasterxyz/snapchain/pull/639/files#diff-2ded749e521d83d7e5fc4bb38f9999b198c03eaa874c47f99cfc6612c38e994bR1161. I'd prefer to leave it in there if possible because there's some annoying work to do to move the correct diffs into this feature. We can merge both features together, but also we're not making other changes to this particular part of the code so it's probably ok to merge this and then merge the next feature with the test shortly after.

src/version/version.rs
```diff
@@ -144,7 +145,8 @@ impl EngineVersion {
             | ProtocolFeature::PrimaryAddresses => self >= &EngineVersion::V5,
             ProtocolFeature::FutureTimestampValidation => self >= &EngineVersion::V6,
             ProtocolFeature::DependentMessagesInBulkSubmit => self >= &EngineVersion::V7,
-            ProtocolFeature::DecoupleShardZeroBlockProduction => self >= &EngineVersion::V8,
+            ProtocolFeature::DecoupleShardZeroBlockProduction
+            | ProtocolFeature::WriteDataToShardZero => self >= &EngineVersion::V8,
```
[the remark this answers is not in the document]
aditiharini: I think with the current implementation old blocks should just pass validation. Empty events, empty txns, empty events hash will pass. 

But being conservative here is probably just worth it. I'll push the V8 release up to 9/3, then put this under V9. which I'll schedule for 9/10.

sanjayprabhu (dismissed):
Great work

PR #637, opened 2025-08-15
feat: Add ability to batch insert into trie

adityapk00:
Add the ability to batch insert into the merkle trie. This will be used by the replicator code

### DIFF
--- a/src/storage/store/engine.rs
+++ b/src/storage/store/engine.rs
@@ -1194,78 +1194,45 @@ impl ShardEngine {
         Ok(events)
     }
 
-    fn update_trie(
-        &mut self,
-        ctx: &merkle_trie::Context,
+    // Compute the keys that need to be updated in the trie. Returns (inserts, deletes)
+    fn get_trie_update_keys(
+        &self,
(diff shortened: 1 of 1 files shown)


src/storage/store/engine.rs
```diff
@@ -1303,7 +1260,88 @@ impl ShardEngine {
...
+    ) -> Result<(), EngineError> {
+        let now = std::time::Instant::now();
+
+        let mut all_inserts = HashSet::new();
+        let mut all_deletes = HashSet::new();
+
+        for event in events {
+            let (inserts, deletes) = self.get_trie_update_keys(event)?;
+
+            // Merge the inserts and deletes
```
sanjayprabhu, line 1308: This looks fine, but might we worth adding a test with a few scenarios unless it's covered by replication tests?

sanjayprabhu (approved):

src/storage/store/engine.rs
adityapk00: The replication tests will cover it (All the replication code will go through this, and there are tests in the other PR that cover this function)

PR #639, opened 2025-08-15
feature: read block events from shard 0 in other shards

aditiharini:


src/mempool/block_receiver.rs
```diff
@@ -0,0 +1,201 @@
...
+            events_hasher.update(&event.hash);
+        }
+
+        if block.header.as_ref().unwrap().events_hash
+            != events_hasher.finalize().as_bytes().to_vec()
+        {
+            return false;
+        }
+
+        // TODO(aditi): Validate signatures
```
sanjayprabhu, line 58: It was not designed for this, so might be annoying to refactor but we should see if we can re-use the code from the block proposer. This logic is important enough that we shouldn't risk introducing divergences by duplicating the code

src/mempool/block_receiver.rs
```diff
@@ -0,0 +1,201 @@
...
+                    None => 1,
+                    Some(last_event) => last_event.block_number(),
+                };
+                self.sync_missing_blocks(
+                    last_block_number,
+                    block.header.as_ref().unwrap().height.unwrap().block_number - 1,
+                )
+                .await;
+                if let Err(BlockReceiverError::ConfirmationTimedOut) = self
+                    .wait_for_confirmation(
```
sanjayprabhu, line 174: I'm not sure this is safe? If we only call recv on events_rx periodically, then the events broadcast channel could fill up and cause us to miss events (and potentially interfere with other subscribers? I need to read up on the implementation).

I think we may need to be constantly listening to events messages in the background and keep tracking the max seq num per shard.

aditiharini: I think this behavior works. It won't interfere with other subscribers. The way the channel works is that it keeps the last n messages and if there's a slow subscriber it drops the oldest message in the buffer and that subscriber will miss events. 

These events should be produced at a very regular 1s cadence so it's fine if some events are dropped because the next one will show up with the latest event seqnum.

src/storage/store/mempool_poller.rs
```diff
@@ -123,14 +132,33 @@ impl MempoolPoller {
...
+                            block_event: None,
+                        })
+                    }
+                    MempoolMessage::OnchainEvent(onchain_event) => {
+                        transaction.system_messages.push(ValidatorMessage {
+                            on_chain_event: Some(onchain_event.clone()),
+                            fname_transfer: None,
+                            block_event: None,
+                        })
+                    }
```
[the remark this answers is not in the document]
aditiharini: Added a comment about this. Will do in a follow up.

sanjayprabhu (approved):
Looks good after fixing conflicts and the adding the metric

src/storage/store/engine.rs
```diff
@@ -536,6 +550,7 @@ impl ShardEngine {
         let mut revoked_signers = HashSet::new();
 
         let mut validation_errors = vec![];
+        let mut last_block_event_seqnum = self.stores.block_event_store.max_seqnum()?;
```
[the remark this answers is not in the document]
aditiharini: Yeah good point. This worked for now because all heartbeats are associated with fid 0 but isn't future proof. Plumbing this value around is annoying :/

src/mempool/block_receiver.rs
aditiharini: Good point, reused.

src/mempool/block_receiver.rs
```diff
@@ -0,0 +1,190 @@
...
+            // The db is the source of truth, it's possible to read this out of the events_rx channel but delivery over that channel is not reliable (it's a broadcast channel) we may not have the most up to date state.
+            let last_stored_event_seqnum = self.stores.block_event_store.max_seqnum().unwrap();
+            let last_event_in_block = block.events.last().unwrap();
+            if last_event_in_block.seqnum() < last_stored_event_seqnum {
+                continue;
+            }
+
+            let first_event_in_block = block.events.first().unwrap();
+            if first_event_in_block.seqnum() > last_stored_event_seqnum + 1 {
+                self.sync_missing_block_events(
```
aditiharini, line 157: We should add some batching here so that initial sync or even later sync if big doesn't throw a bunch of messages in the mempool at the same time. It's also hard to set a reasonable confirmation timeout without some batching.

src/mempool/block_receiver.rs
```diff
@@ -0,0 +1,190 @@
...
+                        seqnum = first_event_in_block.seqnum() - 1,
+                        "Timed out waiting for confirmation",
+                    );
+                    continue;
+                }
+            };
+            self.submit_block(&block).await;
+            // If confirmation fails, we'll try move onto the next block and retry this block if needed.
+            if let Err(BlockReceiverError::ConfirmationTimedOut) = self
+                .wait_for_confirmation(last_event_in_block.seqnum(), Duration::from_secs(1))
```
aditiharini, line 180: These confirmation timeouts should be configurable so they're easy to change. Don't have high conviction in what the numbers should be.

src/mempool/mempool.rs
```diff
@@ -181,23 +181,35 @@ impl proto::Message {
...
+                fname_transfer.proof.as_ref().unwrap().timestamp,
+                fname_transfer.id.to_string(),
+            )
+        } else if let Some(block_event) = &self.block_event {
+            MempoolKey::new(
+                MempoolMessageKind::ValidatorMessage,
+                block_event.block_timestamp(),
+                format!(
+                    "{}-{}",
+                    block_event.block_number(),
```
aditiharini, line 202: We should just use seqnum here. It's unique.

aditiharini: Done

sanjayprabhu (approved):
Looks good. I think there's a few more small improvements we can make but let's save it for a follow up and merge this in.

PR #645, opened 2025-08-19
Add a method to return a FID by an authorized verification address

mehcode:
Implements https://github.com/farcasterxyz/protocol/pull/244

---

> Add `GetFidByVerification` to Hub which allows a developer to take an authorized address and return a FID. 

> The use case here is we ( Calaxy - https://calaxy.com ) are integrating with both Base.app and Farcaster. When logging in with Base, an ethereum address is returned ( ref: https://docs.base.org/base-account/overview/what-is-base-account ). When logging in with Farcaster, a FID is returned. 

> - With an FID, it's possible currently to get a list of authorized ethereum addresses and match a previous "Base.app login" with a new "Farcaster login" to return the same user context

> - With just the address, its not currently possible to return an associated FID

> This will be especially important as current Base.app applications with "Login with Base.app" end up becoming mini apps and using the embedded Farcaster SDK.

> Open to anything with regards to naming. I tried to make it match the feel of what's there.

### DIFF
--- a/src/network/http_server.rs
+++ b/src/network/http_server.rs
@@ -427,6 +427,25 @@ impl FidRequest {
     }
 }
 
+#[derive(Debug, Clone, Deserialize, Serialize)]
+pub struct VerificationFidRequest {
+    #[serde(with = "serdehex")]
+    pub address: Vec<u8>,
+}
+
+impl VerificationFidRequest {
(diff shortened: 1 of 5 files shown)


sanjayprabhu:
Nice. Could you please add store and server level tests?

PR #646, opened 2025-08-19
Update rocksdb (previously pinned ref included in a release)

mehcode:
The currently pinned commit rev was included in release 0.23 of `rocksdb`. Assuming this passes CI, this seems safe to merge.

I updated this all the way to 0.24 as 0.23 does not build on my fresh Arch linux machine (build issues with 0.20 llvm/clang).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1738,9 +1738,9 @@ dependencies = [
 
 [[package]]
 name = "bindgen"
-version = "0.71.1"
+version = "0.72.0"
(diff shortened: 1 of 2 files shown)


sanjayprabhu:
Looks like this was the original reason for pinning this version https://github.com/farcasterxyz/snapchain/pull/16#discussion_r1841294520. 

Thanks. This requires some testing to make sure everything is backwards compatible. We'll merge when we have some time to verify.

PR #648, opened 2025-08-20
feat: Allow conflict-free stores for replicator

adityapk00:
Allow stores to be created in a "conflict-free" mode, which allows duplicates to be inserted (overwritten) and allows conflicting messages with link_compact. The merging does not return an error, and the DB keys are overwritten and re-inserted into the Trie. Some "conflicting" messages - messages that are valid, but would be detected as either prunable or duplilcates are allowed (eg. Duplicate ID Register events). 

These messages are rare, but during replication we accept duplicate or prunable messages since they are in the Merkle Trie.


aditiharini:
LGTM

PR #649, opened 2025-08-20
feat: Server side changes for replicator

adityapk00:
Server side changes for replicator

1. Allow client to specify sort order
2. Return ShardChunk and Block at the snapshot height so that the client can boostrap from that height
3. Return the heighest_fid to expect on each shard, makes client syncing easier.

### DIFF
--- a/src/replication/replication_server.rs
+++ b/src/replication/replication_server.rs
@@ -95,12 +95,6 @@ impl proto::replication_service_server::ReplicationService for ReplicationServer
                         .get_shard_chunk_by_height(request.shard_id, height)
                         .map_err(|e| {
                             Status::internal(format!("Failed to get shard chunk by height: {}", e))
-                        })?
-                        .ok_or_else(|| {
-                            Status::not_found(format!(
-                                "ShardChunk not found for shard {} at height: {}",
-                                request.shard_id, height
-                            ))
                         })?;
 
                     snapshots.push(proto::ShardSnapshotMetadata {
@@ -109,7 +103,7 @@ impl proto::replication_service_server::ReplicationService for ReplicationServer
                         timestamp,
                         highest_fid,
                         block,
-                        shard_chunk: Some(shard_chunk),
+                        shard_chunk,
                     });
                 }
                 snapshots
--- a/src/replication/replication_server.rs
+++ b/src/replication/replication_server.rs
@@ -3,27 +3,65 @@ use std::sync::Arc;
 use tonic::{Request, Response, Status};
 
 use crate::mempool::routing;
-use crate::proto;
+use crate::proto::{self, MessageType, OnChainEventType};
 use crate::replication::replicator::Replicator;
+use crate::storage::store::block::BlockStore;
+
+/// Extract the sort order Enums from the protos
+fn get_sort_order_from_request(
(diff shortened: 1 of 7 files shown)


src/replication/replication_server.rs
```diff
@@ -37,14 +75,45 @@ impl proto::replication_service_server::ReplicationService for ReplicationServer
...
+                    // Get the highest FID for this shard at this height
+                    let highest_fid = self
+                        .replicator
+                        .get_highest_fid_for_shard(request.shard_id, height)
+                        .unwrap_or(0);
+
+                    // Fetch the block for the given height
+                    let block = self.block_store.get_block_by_height(height).map_err(|e| {
+                        Status::internal(format!("Failed to get block by height: {}", e))
+                    })?;
```
sanjayprabhu, line 90: Why do we need the block?

src/storage/store/account/onchain_event_store.rs
```diff
@@ -590,6 +590,45 @@ impl OnchainEventStore {
         Ok((fids, next_page_token))
     }
 
+    pub fn get_highest_fid(&self) -> Result<Option<u64>, OnchainEventStorageError> {
+        let start_prefix = make_onchain_event_type_prefix(OnChainEventType::EventTypeIdRegister);
+        let stop_prefix = increment_vec_u8(&start_prefix);
```
sanjayprabhu, line 595: Do we need to decrement because we're paging in reverse? Or does the pagination logic automatically handle flipping these? Either way, maybe worth a comment.

src/replication/replication_server.rs
```diff
@@ -3,27 +3,65 @@ use std::sync::Arc;
...
-use crate::proto;
+use crate::proto::{self, MessageType, OnChainEventType};
 use crate::replication::replicator::Replicator;
+use crate::storage::store::block::BlockStore;
+
+/// Extract the sort order Enums from the protos
+fn get_sort_order_from_request(
+    system_message_types: Option<proto::SortOrderTypes>,
+    user_message_types: Option<proto::SortOrderTypes>,
+) -> Result<(Vec<OnChainEventType>, Vec<proto::MessageType>), Status> {
```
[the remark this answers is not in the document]
adityapk00: In replicator::build_validator_messages, the ENS proofs and fname proofs are added after the system messages. We  fetch the fnames specially (from the trie, because of the duplicate fnames issue), so its not in the sort order

src/replication/replication_server.rs
adityapk00: I thought we should insert the block into shard 0 as well, so that all shards are at the same block height and can start syncing. Let me know if you think this is not a good idea, and we should leave shard 0 empty

src/storage/store/account/onchain_event_store.rs
adityapk00: Yes, for paging in reverse. I'll add a comment

src/replication/replication_server.rs
adityapk00: Convo summary: We'll get the latest height for each shard separately.

adityapk00:
Closing this PR. Switching to trie-based iteration, so all this code is no longer needed.

PR #650, opened 2025-08-22
fix: correctly route fnames to all shards

aditiharini:
We had an off-by-one bug where we're only routing to shard 1 in addition the desired shard for fnames and never routing to shard 2 in addition. 

We don't need to put this behind a protocol version upgrade because it's not changing consensus-critical logic. 
(1) If the proposer has been upgraded, then it will route fnames to all shards and all nodes will end up these fnames in all shards. 
(2) If the proposer has not been upgraded, then it will route fnames to only some shards (incorrectly as is happening now) and nodes will end up with these fnames in whatever shards they were routed to. 

Adding a new flag is not totally free because we will have to maintain these flags until we have block pruning enabled on all nodes and it makes the code significantly harder to reason about.


sanjayprabhu (approved):
Nice, thank you.

PR #651, opened 2025-08-22
feat: Allow attaching orphaned nodes to the merkle trie

adityapk00:
When we write to the trie from parallel threads (like we do in the replicator), we will write "orphaned" nodes. These are nodes that are written to the DB, but not yet attached to the root node. This is because we want to write in parallel threads to different parts of the trie. 

Once the operation is done, we call `attach_to_root` and `recalculate_hashes` to add new nodes to the trie. This way, we can parallelize construction of the trie. 

Also remove excluded_hash() and snapshot() which we don't use anymore. The `get_trie_node_metadata` is still supported, which can be used to get info about the parts of a trie.

### DIFF
--- a/src/storage/trie/merkle_trie_tests.rs
+++ b/src/storage/trie/merkle_trie_tests.rs
@@ -3,6 +3,7 @@ mod tests {
     use crate::storage::db::{RocksDB, RocksDbTransactionBatch};
     use crate::storage::store::account::IntoU8;
     use crate::storage::trie::merkle_trie::{Context, MerkleTrie, TrieKey};
+    use crate::storage::trie::util;
     use crate::utils::factory::{events_factory, messages_factory};
 
     fn random_hash() -> Vec<u8> {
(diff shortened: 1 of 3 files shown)


src/storage/trie/merkle_trie_tests.rs
```diff
@@ -138,4 +139,219 @@ mod tests {
...
+        assert_eq!(trie.root_hash().unwrap(), root_hash);
+
+        // Key for the node to mess up and then re attach
+        let key = &key_vecs[0][0..6]; // try to attach this node
+        let xkey = (key_conv.expand)(key); // with this expanded key
+        let xprefix = &xkey[0..xkey.len() - 1]; // parent's prefix
+        let child_char = xkey[xkey.len() - 1]; // child's character in the parent's children[]
+
+        println!(
+            "Key[0] is {:?}, xkey is {:?}",
```
sanjayprabhu, line 180: Delete?

src/storage/trie/merkle_trie_tests.rs
```diff
@@ -138,4 +139,219 @@ mod tests {
...
+
+        let tmp_path1 = tempfile::tempdir().unwrap();
+        let db1 = &RocksDB::new(tmp_path1.path().to_str().unwrap());
+        db1.open().unwrap();
+
+        let tmp_path2 = tempfile::tempdir().unwrap();
+        let db2 = &RocksDB::new(tmp_path2.path().to_str().unwrap());
+        db2.open().unwrap();
+
+        let branching_factor = 16;
```
sanjayprabhu, line 285: We should just hardcode this into the trie at this point. We cannot change it anymore without breaking roots.

sanjayprabhu (approved):

PR #652, opened 2025-08-25
fix: Paged iterator doesn't go through sub-keys

adityapk00:
The paged iterator needs to add [0] to the page_token so that it captures any sub-keys, instead of increment_u8 which will skip over all sub keys during iteration

### DIFF
--- a/src/storage/db/rocksdb.rs
+++ b/src/storage/db/rocksdb.rs
@@ -366,7 +366,7 @@ impl RocksDB {
             start_iterator_prefix
         } else {
             if let Some(page_token) = &page_options.page_token {
-                increment_vec_u8(&page_token)
+                [page_token.clone(), vec![0u8]].concat()
             } else {
                 start_iterator_prefix
             }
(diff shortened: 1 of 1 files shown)


src/storage/db/rocksdb.rs
```diff
@@ -366,7 +366,7 @@ impl RocksDB {
             start_iterator_prefix
         } else {
             if let Some(page_token) = &page_options.page_token {
-                increment_vec_u8(&page_token)
+                [page_token.clone(), vec![0u8]].concat()
```
sanjayprabhu, line 369: Could you please add a comment with an example subkey? This might be hard to follow for someone without context.

sanjayprabhu (approved):

src/storage/db/rocksdb.rs
adityapk00: Done

sanjayprabhu (approved):

PR #654, opened 2025-08-26
Add GNU GPL v3 license

sanjayprabhu:
Add GNU GPL v3 license to Snapchain.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,674 @@
+                    GNU GENERAL PUBLIC LICENSE
+                       Version 3, 29 June 2007
+
+ Copyright (C) 2007 Free Software Foundation, Inc. <https://fsf.org/>
+ Everyone is permitted to copy and distribute verbatim copies
+ of this license document, but changing it is not allowed.
+
+                            Preamble
+
+  The GNU General Public License is a free, copyleft license for
+software and other kinds of works.
+
+  The licenses for most software and other practical works are designed
+to take away your freedom to share and change the works.  By contrast,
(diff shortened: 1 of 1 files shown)


nuconomy:
+1

andrei0x309:
While a permissive license might be better for greater involvement, especially that which is meant to creating commercial products. 

A copyleft license is better for open-source consolidation if this project becomes very popular.

So this looks like a bet. At this point, it depends on whether Farcaster pans a long-term investment in the project.

If the investment plan is longer than 4 years, I would pick a copy left license.

byte:
I've added my 2 cents - https://farcaster.xyz/bytebot/0x43001273

I do believe the GPLv2/v3 are ideal, eventhough the AGPL might fit better (but isn't my choice). Permissive licenses are great, but I do not think they will benefit the protocol overall if people fork and run with something new, anyway.

shazow:
> A copyleft license is better for open-source consolidation if this project becomes very popular.

There is mountains of evidence that the opposite is true in 2025. Copyleft licenses lead to fragmentation and a bunch of competing standards, and typically the permissive ones win. I share several examples in my post.

daferna:
I'll add that ATProto is currently dual licensed under MIT & Apache2, since Apache 2 apparently has some extra software patent protection vagaries that some care about.
I think Snapchain at the very least should match the licenses of it peer competitors, and as such I can't support copyleft licenses here.

varunsrin:
Thanks appreciate the feedback here and on the community calls. We made the decision to go ahead with the proposal for GNU GPL v3.

PR #658, opened 2025-08-28
WIP feat: periodically revoke invalid username proofs

aditiharini:


manan19:
Closing in favor of a phased rollout — full plan and open questions tracked in #852.

Summary of why this needs to be re-scoped rather than landed as-is:
- `Stores::revoke_message` (singular) called from `ShardEngine::revalidate_user_message` does not exist on `main` — only `revoke_messages` (plural, signer-scoped) does, so this would not compile against current `main`.
- Revoking the `UserNameProof` does not cascade to the linked `UserDataAdd::Username`, so stale display names would persist after revocation. Precedent for the cascade lives at `src/storage/store/engine.rs` ~L837–L868 (fname-transfer path).
- `ValidatorMessage::mempool_key` for the new `RevalidateMessage` variant has an unresolved `TODO` and currently hardcodes shard `0`.
- Consensus determinism around accepting an off-chain RPC result (`external_data.ens_resolved_address`) needs an explicit decision before validators act on it.

Phase 1 in #852 ships the detection job (read-only, observability) so we can measure the problem and the RPC budget before introducing any state mutation.

PR #663, opened 2025-09-04
feat: Add trie-based replication server

adityapk00:
Modify the replication server to use trie-based iteration to get all messages instead of by DB.


src/replication/replicator.rs
```diff
@@ -73,97 +202,408 @@ impl Replicator {
...
+            let fid = decoded_key.fid;
+            let onchain_message_type = decoded_key.onchain_message_type;
+            let message_type = decoded_key.message_type;
+            let rest = decoded_key.rest;
+
+            fids_in_page.insert(fid);
+            match (onchain_message_type, message_type) {
+                (Some(onchain_event_type), None) => {
+                    // On-chain event
+                    if onchain_event_type == 7 {
```
sanjayprabhu, line 430: We should make this a constant.

src/proto/replication.proto
```diff
@@ -20,27 +18,66 @@ message ShardSnapshotMetadata {
...
+  uint64 fid = 1;
+  bytes account_root_hash = 2;
+  uint64 num_messages = 3;
+}
+
+message ReplicationTriePartStatus {
+  uint32 shard_id = 1;
+  uint64 height = 2;
+  uint32 virtual_trie_shard = 3;
+  uint32 last_response = 4;
```
sanjayprabhu, line 78: What is this used for?

src/replication/replication_server.rs
```diff
@@ -36,15 +26,52 @@ impl proto::replication_service_server::ReplicationService for ReplicationServer
...
-                    shard_id: request.shard_id,
-                    height,
-                    timestamp,
-                })
-                .collect(),
+        // We store the snapshots by the data shards, so there's no snapshot for shard-0. That's OK,
+        // because for shard-0, we just need the highest block number, so get it from any shard,
+        // and return the actual shard-0 block
+        let get_shard_id = if request.shard_id == 0 {
+            1
```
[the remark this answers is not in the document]
adityapk00: So what's happening here is that if the request is for shard 0, we don't have a snapshot available for shard-0 (Snapshots are created only for shard 1/2). So, if the user requests the metadata for shard-0, we actually need to go to the live block_store (not the snapshot) and get the latest block from there. 

To make this less confusing, I'm now explicitly handling it, so the code is clear that this is what is happening.

src/replication/replicator.rs
```diff
@@ -1,21 +1,148 @@
...
+    cache_size: usize,
+
+    // Key (shard, fid, message_type) -> HashMap<hash, Message>
+    cache: HashMap<MessageHashCacheKey, Arc<HashMap<Vec<u8>, CacheMessageEntry>>>,
+
+    // Key (shard, fid, message_type) -> Last used timestamp
+    cache_lru: HashMap<MessageHashCacheKey, SystemTime>,
+}
+
+impl MessageHashCache {
```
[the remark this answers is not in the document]
adityapk00: added size, hit, miss stats

src/replication/replicator.rs
adityapk00: Done

PR #665, opened 2025-09-05
feat: support storage lending in the shard and block engines

aditiharini:
Support storage lending. 

This feature enables support for storage lending in the `BlockEngine` and `ShardEngine` but the messages aren't handled properly in the rpc server and mempool yet.

### DIFF
--- a/src/storage/store/engine.rs
+++ b/src/storage/store/engine.rs
@@ -1241,6 +1241,10 @@ impl ShardEngine {
                     Ok(vec![])
                 }
             }
+            MessageType::LendStorage => {
+                // Pruning not supported for storage lend messages
+                Ok(vec![])
+            }
             unhandled_type => {
                 return Err(EngineError::UnsupportedMessageType(unhandled_type));
             }


src/proto/blocks.proto
```diff
@@ -101,20 +101,26 @@ message ConsensusMessage {
...
 enum BlockEventType {
-  BLOCK_EVENT_TYPE_HEARTBEAT = 0; 
+  BLOCK_EVENT_TYPE_HEARTBEAT = 0;
+  BLOCK_EVENT_TYPE_LEND_STORAGE = 1;
 }
 
 message HeartbeatEventBody {
 }
 
+message LendStorageEventBody {
```
aditiharini, line 111: We may want to make this a more general `MergeMessage` event.

src/replication/replicator.rs
```diff
@@ -376,6 +376,12 @@ impl Replicator {
                 )?
                 .messages
             }
+            proto::MessageType::LendStorage => {
+                stores
```
aditiharini, line 380: @adityapk00 do these messages go through the engine validations? I added a validation to make sure that storage lends can't be merged into shard 1/2 directly.

adityapk00: No, the replicator client is not doing message validations right now. (There's 1000s of messages failing validations right now for various reasons) Right now, we only do hash and signature check

src/storage/store/block_engine.rs
```diff
@@ -132,34 +200,180 @@ impl BlockEngine {
...
                 }
             }
         }
 
+        let mut storage_slot = self
+            .storage_slot_for_transaction(snapchain_txn, false)
+            .unwrap();
+
+        for message in &snapchain_txn.user_messages {
+            match self.validate_user_message(message, &storage_slot, timestamp, version, txn_batch)
```
sanjayprabhu, line 337: We also need to account for the user hitting the limits on the storage lend message itself, right? i.e. there's no pruning of storage lend messages happening right now.

And when prunes happen, we need to account for the deleted lend messages and emit block events for those.

Practically, I'm not sure if we'll hit these limits, but we can't allow infinite storage lend messages.

src/storage/store/engine.rs
```diff
@@ -785,7 +784,44 @@ impl ShardEngine {
...
+                                    &lend_storage_event.lend_storage_message
+                                {
+                                    match self
+                                        .stores
+                                        .storage_lend_store
+                                        .merge(lend_storage_message, txn_batch)
+                                    {
+                                        Ok(hub_event) => {
+                                            merged_messages_count += 1;
+                                            self.update_trie(trie_ctx, &hub_event, txn_batch)?;
```
sanjayprabhu, line 805: We'd be inserting this into the trie for the lender fid right? That seems wrong? I think it's more correct to insert into the borrower fid as a record of have received the lent storage.

src/proto/message.proto
```diff
@@ -210,4 +213,16 @@ message FrameActionBody {
...
+}
+
+enum StorageUnitType {
+  UNIT_TYPE_LEGACY = 0;
+  UNIT_TYPE_2024 = 1;
+  UNIT_TYPE_2025 = 2;
+}
+
+message LendStorageBody {
+  uint64 to_fid = 1;
```
sanjayprabhu, line 225: We don't need the from fid? Might be useful to have it just in case, esp.  if this is the same body that will be emitted as an event. I can see wanting to know who the storage came from.

PR #666, opened 2025-09-05
feat: Add profile token user data type

sanjayprabhu:
Implement https://github.com/farcasterxyz/protocol/discussions/247. 

Will go live on the protocol on Sep 24 (engine version 10)


aditiharini:
Should add a test case in client-parity-tests and update the client lib to have the equivalent validation.

PR #667, opened 2025-09-05
feat: Add the ability to start a node via replication

adityapk00:
These are the client changes to support replication. It's not enabled yet, once we feel confident with some manual testing, I'll turn it on.


src/bootstrap/service.rs
```diff
@@ -0,0 +1,1285 @@
...
+    shutdown: Arc<AtomicBool>,
+    network: crate::proto::FarcasterNetwork,
+    statsd_client: StatsdClientWrapper,
+    shard_ids: Vec<u32>,
+    rocksdb_dir: String,
+}
+
+impl ReplicatorBootstrap {
+    // Postfixes used to store progress data into the DB
+    const WORK_UNIT_POSTFIX: u8 = 1u8;
```
aditiharini, line 94: We should use the store in node_local_state.rs for storing progress on this.

src/bootstrap/service.rs
```diff
@@ -0,0 +1,1285 @@
...
+                    error!("Failed to send DB commit response from DB Writer thread");
+                    break;
+                }
+            }
+
+            info!("DB writer thread finished for shard {}", shard_id);
+        })
+    }
+
+    /// Create a dedicated thread (per shard) to write to the Trie. This way, we can keep all the trie nodes
```
aditiharini, line 500: Is there churn in what's kept in memory if there's only 1 thread because the trie nodes get kicked out of memory when the account stores are written?

src/bootstrap/replication_client_test.rs
```diff
@@ -0,0 +1,206 @@
...
+            trie::merkle_trie::MerkleTrie,
+        },
+        utils::factory::{self, messages_factory},
+    };
+    use std::time::Duration;
+    use tempfile::TempDir;
+    use tonic::transport::Server;
+
+    #[tokio::test]
+    async fn test_replication_client_bootstrap() {
```
aditiharini, line 26: Is there any scenario where we may reprocess messages and trie keys after they've already been committed? If so, can we add a test to make sure the deduplication logic works properly?

adityapk00:
I'm working on the rest of the comments, but here's a quick table of all the perf gains I got. I did try a bunch of other stuff, but stuff that didn't work is not included here. The reason to focus so much on client performance is that the client needs to finish the replication before the server expires the snapshots (currently set to 1 day). We can't increase the server snapshot expiry too much either, because the server's RocksDB will need to hold the snapshot open, which slows down subsequent writes to the DB. Plus, if the client doesn't finish the replication before the server expires the snapshot, it can't complete it at all, and will need to restart the whole thing from scratch. 

Description | Messages per second
-- | --
Base | 6k /sec
WAL off | 7 k/sec
Bulk write (buf size =512MB parallel=8 max_write = 4 min_buffers_to_merge=2) | 7.4 k/sec
batched DB commit (2k messages | 12.9 k/sec
batched trie insert per FID | 14.4 k/sec
Separate out DBWriter thread | 36 k/sec
conflict-free-store, Don't write events | 44 k/sec
1 thread per virtual trie shard | 67 k/sec
Close/Open DB every 5m messages | 88 k/sec
Separate out trie thread, don't post-process | 92 k/sec
Separate out fetching thread, (prefetch_batch_count = 2) | 108 k/sec
do batch_verify for signatures | 142 k/sec

src/bootstrap/service.rs
adityapk00: So the trie does not get kicked out of memory after the account stores are written, because this thread keeps its own copy of the trie in memory. This is where most of the perf gain comes from. By keeping most of the top layers of the trie in memory, we can avoid upto 12 DB reads per insert, speeding up things quite a bit. This trie is reloaded after the db batch of 5 million messages. 

In any case, the main reasons to do a trie in its own thread is so that we don't have to do post-processing anymore. The trie root hash is always valid since there are no competing threads. I'll update the comment to reflect this.

src/bootstrap/replication_client_test.rs
adityapk00: Yes, it does handle duplicates properly, but I will add a test of this. Actually, there are other scenarios we should test, like an interrupted connection etc... I'll add these tests in the next PR, since I have some test utility classes there. Your comment about the testing is a good one - We should have more extensive tests, since there are many edge cases that get handled. I'll add all these into the next PR

src/bootstrap/service.rs
```diff
@@ -0,0 +1,1285 @@
...
+        )
+        .await
+        .map_err(|e| BootstrapError::DatabaseError(e.to_string()))?;
+
+        // Buffer size is 2, so it will prefetch the next page without waiting
+        let (messages_page_tx, mut messages_page_rx) = tokio::sync::mpsc::channel(2);
+
+        let rpc_shutdown_signal = Arc::new(AtomicBool::new(false));
+        let rpc_shutdown_signal_clone = rpc_shutdown_signal.clone();
+        let rpc_client_handle = tokio::spawn(async move {
```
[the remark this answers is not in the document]
aditiharini: Makes sense

PR #671, opened 2025-09-11
fix: Simplify the replication by removing trie/DB threads

adityapk00:
Simplify the replicator code by removing the trie / DB threads. Instead, we'll use tokio's Mutex() to lock the DB and Trie (separately) while processing. 

This is effectively the same thing, but is simpler to reason about because we can just use `.await` everywhere instead of using tokio threads and tx/rx channels to communicate. Makes the code simpler. This PR also separates out the "core" logic of commit to DB, commit to Trie from the parallel threads orchestration to make the code easier to reason about.

### DIFF
--- a/src/bootstrap/service.rs
+++ b/src/bootstrap/service.rs
@@ -863,6 +863,14 @@ impl ReplicatorBootstrap {
         let db = thread_engine.get_stores().db.clone();
 
         for fid in &fids_to_check {
+            let actual_vts = TrieKey::fid_shard(*fid);
+            if actual_vts != virtual_trie_shard {
+                return Err(BootstrapError::GenericError(format!(
+                    "Fid {} was found in vts {} but belongs to vts {}",
+                    fid, virtual_trie_shard, actual_vts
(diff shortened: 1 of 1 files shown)


src/bootstrap/service.rs
```diff
@@ -54,30 +58,35 @@ pub enum WorkUnitResponse {
...
     peer_address: String,
     shutdown_signal: Arc<AtomicBool>,
-    db_commit_tx: mpsc::Sender<DbCommitRequest>,
-    trie_insert_tx: mpsc::Sender<TrieInsertRequest>,
+    thread_engine: Arc<ShardEngine>,
+}
+
+// Shared data for all threads working on a shard. The DB and the Trie
+// are shared among all the vts.
+struct ShardSharedWorkData {
```
aditiharini, line 78: Do we want all these things to be separately lockable? Or do we want 1 lock over the whole thing?

aditiharini:
Looks good, I like this change a lot.

PR #672, opened 2025-09-12
feat: Add client manager + tests for replication rpc requests

adityapk00:
- Separate out the client RPC into it's own class to simplify replicator logic
- Tests for client RPC manager
- Optionally allow specifying which peers to replicate against


sanjayprabhu (dismissed):
Nice

PR #674, opened 2025-09-13
fix: inserting duplicate keys in the trie fails whole transaction

adityapk00:
There's an edge case where having duplicate keys in insert() was causing the whole transaction to fail to add. Fix + tests


aditiharini:
Could you add an explanation of what the bug was in the existing code in the PR description? I'm having a hard time reasoning about it based on the code.

aditiharini:
LGTM

adityapk00:
Closing this in favor of #679

PR #682, opened 2025-09-18
feat: Discover nodes to sync with using gossip p2p

adityapk00:
when replication starts, we seed it with an initial bootstrap list. The replication service then spins up a lightweight gossip node to discover peers to sync against.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -255,9 +255,6 @@ fn create_replicator(
 }
 #[tokio::main]
 async fn main() -> Result<(), Box<dyn Error>> {
-    crypto::CryptoProvider::install_default(ring::default_provider())
-        .expect("Failed to install rustls crypto provider");
-
     let args: Vec<String> = std::env::args().collect();
 
     let app_config = match snapchain::cfg::load_and_merge_config(args) {
@@ -330,6 +327,10 @@ async fn main() -> Result<(), Box<dyn Error>> {
     if db_is_empty {
         match app_config.snapshot.bootstrap_method {
             BootstrapMethod::Replicate => {
+                // Initialize SSL for rustls
+                crypto::CryptoProvider::install_default(ring::default_provider())
+                    .expect("Failed to install rustls crypto provider");
+
                 // info!("Starting node with replication bootstrap");
                 // let replicator = ReplicatorBootstrap::new(statsd_client.clone(), &app_config);
 


src/bootstrap/replication/peer_discovery.rs
```diff
@@ -0,0 +1,161 @@
...
+            if ts.elapsed() < INCOMPATIBLE_PEER_COOLDOWN {
+                return;
+            }
+        }
+
+        // Spawn validation (metadata height match) via RpcClientsManager::add_new_peer
+        let rpc_manager = self.rpc_manager.clone();
+        let incompatible = self.incompatible_peers.clone();
+        info!(peer = %peer_id, addr = %http_addr, "Discovered potential replication peer");
+        tokio::spawn(async move {
```
aditiharini, line 148: Why do we need a thread for each of these calls?

aditiharini:
Is there any place where we kick bad peers from the peer manager? is there a need for this?

src/bootstrap/replication/peer_discovery.rs
```diff
@@ -0,0 +1,161 @@
...
+        gossip_config: &GossipConfig,
+        rpc_manager: Arc<RpcClientsManager>,
+        target_height: u64,
+        fc_network: crate::proto::FarcasterNetwork,
+        statsd_client: StatsdClientWrapper,
+        shutdown_rx: tokio::sync::oneshot::Receiver<()>,
+    ) -> Result<Self, BootstrapError> {
+        // Ephemeral keypair for discovery session
+        let keypair = Keypair::generate();
+        let gossip = SnapchainGossip::create(
```
aditiharini, line 44: If we start up a regular gossip system
(1) Will we be advertising our own contact infos? What will happen if peers try to connect to us before initial sync is complete? 
(2) Will we be subscribed to decided values? This is unnecessary but isn't an actual problem, I just want to make sure I understand the behavior of the system.

src/bootstrap/replication/peer_discovery.rs
adityapk00: you're right - we need only one of the spawn. I've removed the one from add_peer
We need to keep this one because I didn't want to make handle_contact_info async and .await a network connection in the gossip messages handler. It is possible that the connection takes multiple seconds/stalls, and I didn't want to block the gossip selector in case the peer connection stalls.

src/bootstrap/replication/peer_discovery.rs
adityapk00: I didn't want to start a full gossip system because we don't have a lot of data that will be asked to serve - Like if we get sync requests for blocks or shard chunks, we don't have them, so we can't serve them/return errors, which might cause peers to blacklist us, and we havent even started yet. 

We dont handle decided values until after replication is done and we start the full gossip system.

src/main.rs
```diff
@@ -249,9 +253,11 @@ fn create_replicator(
     );
     Arc::new(replicator)
 }
-
 #[tokio::main]
 async fn main() -> Result<(), Box<dyn Error>> {
+    crypto::CryptoProvider::install_default(ring::default_provider())
+        .expect("Failed to install rustls crypto provider");
```
sanjayprabhu, line 259: Can this fail? Should we only do this if we need to replicate?

src/bootstrap/replication/peer_discovery.rs
```diff
@@ -0,0 +1,160 @@
...
+        let incompatible = self.incompatible_peers.clone();
+        info!(peer = %peer_id, addr = %http_addr, "Discovered potential replication peer");
+        tokio::spawn(async move {
+            match rpc_manager.add_new_peer(http_addr.clone()).await {
+                Ok(true) => {
+                    info!(peer = %peer_id, addr = %http_addr, "Added new replication peer")
+                }
+                _ => {
+                    warn!(peer = %peer_id, addr = %http_addr, "Peer incompatible or validation failed; cooling down");
+                    incompatible.lock().unwrap().insert(peer_id, Instant::now());
```
sanjayprabhu, line 155: I wonder if it's worth separating unreachable vs doesn't have snapshot. But in practice, I guess it doesn't matter much why it failed?

sanjayprabhu (approved):
Nice, looks good.

src/main.rs
adityapk00: fixed

sanjayprabhu (approved):

PR #683, opened 2025-09-19
feat: implement pruning for storage lends

aditiharini:
Prune revokes for storage lends (messages where num_units is set to 0) as soon as they're merged. We don't want these message to consume storage and we don't expect conflicts to be frequent so the fact that these messages are pruned immediately should not be consequential to users. 

In order to protect against reconciling old messages when these revokes have been pruned, we set the recency window to 1hr for storage lend messages.

### DIFF
--- a/src/storage/store/engine.rs
+++ b/src/storage/store/engine.rs
@@ -854,7 +854,7 @@ impl ShardEngine {
                     {
                         error!(
                             seqnum = block_event.seqnum().to_string(),
-                            "error merging block event: {}",
+                            "Error merging block event: {}",
                             err.to_string()
                         );
                     } else {


src/core/validations/error.rs
```diff
@@ -146,6 +146,8 @@ pub enum ValidationError {
     MissingEmbed,
     #[error("timestamp more than 10 mins in the future")]
     TimestampTooFarInFuture,
+    #[error("timestamp more than 1 hour in the past")]
```
aditiharini, line 149: I'm actually leaning toward making this even tighter, just enough time to account for clock skew. Maybe 10 mins is reasonable.

src/core/validations/message.rs
```diff
@@ -208,6 +208,9 @@ pub fn validate_message(
             validate_frame_action_body(&frame_action_body)?;
         }
         Some(proto::message_data::Body::LendStorageBody(lend_storage_body)) => {
+            if message_data.timestamp < block_timestamp.decr_by(60 * 60).to_u64() as u32 {
```
aditiharini, line 211: Leaning toward making this window even smaller. Maybe something like 10m just to account for clock skew and minor delays.

sanjayprabhu: Sounds good. I think it's easier to keep it more restrictive and relax later. Normal users won't hit this, so should be fine to be aggressive.

sanjayprabhu, line 211: Let's add a comment about why we need this, especially just for StorageLend. I also wonder if this should be more prominent, in the engine vs hidden away within validations.

src/core/validations/message_test.rs
```diff
@@ -676,4 +676,20 @@ mod tests {
         );
         assert_validation_error(&msg, ValidationError::ExceededMaxStorageUnits);
     }
+
+    #[test]
+    fn test_storage_lend_timestamp_too_far_in_past() {
```
sanjayprabhu, line 681: Let's add an explicit block engine test for a user replaying an old storage lend message after lender has zeroed out the lent storage.

sanjayprabhu (commented):
Nice, this is cleaner than I expected.

PR #685, opened 2025-09-19
feat: Ensure we have at least 65k FD limit to start replicator

adityapk00:
On unix systems, we can read the ulimit and even set it (up to the hard limit). We set it to 65k to ensure the replicator runs properly.


sanjayprabhu (dismissed):
Nice

adityapk00:
Updated slightly to use nix instead of libc so we can avoid the unsafe{}

PR #689, opened 2025-09-22
feat: Add progress for replication and other nice to haves

adityapk00:
- Progress with ETA for replication
- timeouts for RPC calls
- Handle the block 0 for devnet/tests
- store shard-0 metadata + stats

### DIFF
--- a/src/bootstrap/replication/client_test.rs
+++ b/src/bootstrap/replication/client_test.rs
@@ -77,7 +77,17 @@ mod tests {
                     Err(status) => Err(status.clone()),
                 }
             } else {
-                Err(Status::not_found("No metadata for this shard_id"))
+                // return empty for shard-0
+                if shard_id == 0 {
+                    return Ok(Response::new(GetShardSnapshotMetadataResponse {
+                        snapshots: vec![ShardSnapshotMetadata::default()],
+                    }));
+                }
+
+                Err(Status::not_found(format!(
+                    "No metadata for shard_id {}",
+                    shard_id
+                )))
             }
         }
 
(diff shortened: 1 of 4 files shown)


src/bootstrap/replication/client_test.rs
```diff
@@ -1227,7 +1236,14 @@ mod tests {
         config.fc_network = source_engine.network.clone();
         config.snapshot.replication_peers = vec![server_addr];
 
-        let bootstrap = ReplicatorBootstrap::new(statsd_client(), &config);
+        let bootstrap = ReplicatorBootstrap::new_with_config(
+            statsd_client(),
+            &config,
+            ReplicatorBootstrapConfig {
+                rpc_retry_delay: Duration::from_secs(0),
```
aditiharini, line 1243: Why is this set to 0?

src/bootstrap/replication/service.rs
```diff
@@ -271,30 +289,42 @@ impl ReplicatorBootstrap {
         // Create tasks for each shard to run in parallel
         let mut shard_tasks = JoinSet::new();
 
-        for shard_id in shard_ids {
-            let rpc_client_manager = match rpc_client_managers.get(&shard_id) {
+        // For shard-0, we'll fetch the latest metadata from the peer. Since we don't need to sync shard-0, we'll just store the metadata
```
aditiharini, line 292: To confirm, this will be set to the first block we should start syncing from (which is not the latest/current block).

src/bootstrap/replication/service.rs
```diff
@@ -1119,64 +1185,64 @@ impl ReplicatorBootstrap {
     ) -> Result<(), BootstrapError> {
         info!("Writing final block and shard chunks to database...");
 
+        // Write shard-0 block
```
aditiharini, line 1188: Why did we need to make this change? What's the difference in behavior and why was it necessary? It's not obvious to me why the existing code is wrong.

src/bootstrap/replication/service.rs
adityapk00: right.

aditiharini (approved):

PR #729, opened 2025-12-04
feat: separate proto files into standalone crate

christopherwxyz:
## Summary
- Extracts protocol buffer definitions into a separate `snapchain-proto` crate so consumers can import protos without depending on the entire snapchain project
- Converts the project to a Cargo workspace with `proto/` as a workspace member
- Uses extension traits pattern to maintain functionality while respecting Rust's orphan rules

## Changes
- Create `proto/` workspace member with its own `Cargo.toml` and `build.rs`
- Move proto definitions from `src/proto/` to `proto/definitions/`
- Move proto-only impls (Height, ShardHash, Message, FullProposal, ConsensusMessage, BlockEvent) to proto crate
- Add extension traits for methods requiring main crate types:
  - `CommitsExt` for commit certificate conversion
  - `FullProposalExt` for proposer address
  - `HubEventExt` for event construction  
  - `HubEventStorageExt` for storage operations
- Rename `HubEvent::from()` to `HubEvent::new_event()` to avoid trait conflict with `std::convert::From`

## Test plan
- [ ] Docker compose build passes
- [ ] All existing tests pass
- [ ] Proto crate can be imported independently


aditiharini:
@christopherwxyz  Was moving the extension traits into the proto crate necessary due to some compilation issue or was it a choice?

aditiharini:
We need to make sure to update the script here to generate protos for the client libraries https://github.com/farcasterxyz/hub-monorepo/blob/main/generate-protos.sh

christopherwxyz:
> @christopherwxyz Was moving the extension traits into the proto crate necessary due to some compilation issue or was it a choice?

I think the extension traits pattern is necessary here due to [Rust's orphan rule](https://doc.rust-lang.org/reference/items/implementations.html). When we extract proto types into a separate crate, we can't implement traits or add inherent methods that depend on types from the main crate (because we'd be implementing foreign traits on foreign types, or need types that don't exist in the proto crate).

Will update the script.

PR #774, opened 2026-03-20
add LoggingConfig with production preset to reduce log volume

shrilakshmikakati:
src/cfg.rs

A new LoggingConfig struct was added with:

level – preset – "default" or "production"; global minimum log level ("info" by default). Hot pathways are silenced by the production preset: network → error, consensus/storage/mempool → warn
Consensus_level, storage_level, network_level, and mempool_level are per-subsystem overrides for more precise control.

A new function called build_env_filter() generates a directive string that is compatible with tracing subscribers.
To enable operations to override the level at startup without modifying the configuration file, the --log-level CLI flag was added.
LoggingConfig was added to the default implementation of the main Config struct.
Four unit tests covering default behavior, production preset, per-subsystem override, and CLI flag override have been added.

src/main.rs

Build_env_filter() was wired into the tracing subscriber setup; the hardcoded "info" fallback was replaced with a single line modification.

### DIFF
--- a/src/cfg.rs
+++ b/src/cfg.rs
@@ -80,9 +80,130 @@ impl Default for ReplicationConfig {
     }
 }
 
+/// Controls verbosity of the node's structured logging output.
+///
+/// Operators running with a log aggregator (Datadog, Grafana Cloud, etc.) should
+/// set `preset = "production"` to suppress high-frequency debug/info logs from
+/// hot-path subsystems (consensus, storage, mempool, network) while still
+/// surfacing warnings and errors from every other module.
+///
+/// Fine-grained per-subsystem overrides are also available and take precedence
+/// over the preset when both are set.
+///
+/// Example TOML:
+/// ```toml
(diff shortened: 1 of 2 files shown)


topocount (changes requested):
@shrilakshmikakati thanks for your patience here! I think this is a bit overkill for what we need. Setting the [RUST_LOG env var](https://rust-lang-nursery.github.io/rust-cookbook/development_tools/debugging/config_log.html) in the default [docker compose](https://github.com/farcasterxyz/snapchain/blob/7bae87664c1f65ec92c6a4295f36e6404097f7d9/docker-compose.mainnet.yml#L9-L10) setup should be the big improvement if we ignore INFO logs all modules exept those related to gossip, so it's clear whether or not one's node is connected.

Secondly, It'd be good to convert the "duplicate" onchain message log errors to info because they aren't really errors.

papabeartaurus420-png (approved):

PR #849, opened 2026-04-27
test: improve diagnostics on two flaky tests

manan19:
## Summary

Two tests have been flaking in CI but their assertion messages are too thin to diagnose the cause when they fail. This PR improves the diagnostics without changing the test logic — the happy path is unchanged; only the failure mode gets richer.

### `cast_test::test_cast_validation`
The current code does `validate_cast_add_body(...).is_ok()` and then prints `{:?}` of the entire cast as the assert message. Two problems:
1. The actual validation error is discarded.
2. Long casts make the message unreadable.

After: capture the error explicitly and print it plus cast metadata (text length, embed/mention counts, type, parent presence). The full cast body stays out of the message.

### `server_tests::test_get_on_chain_signers_by_fid`
Failed in [PR #848 CI](https://github.com/farcasterxyz/snapchain/pull/848) at `assert_eq!(events.len(), 1)` (got 0) on the *paginated* request, while the first request returned 1 event correctly. We don't know whether the missing event was never written to the store or whether pagination dropped it.

After: when the paginated assertion fails, re-query the same fid without pagination and panic with both the paginated event set and the full event set, including types and block numbers — enough to distinguish "store missing the second add" from "pagination dropped it" the next time it flakes.

This PR does not address root cause for either flake; it only makes the next failure diagnosable.

## Test plan

- [x] `cargo check --tests` passes
- [x] `cargo test test_get_on_chain_signers_by_fid` passes locally (happy path unchanged)
- [ ] CI green


manan19:
@copilot resolve the merge conflicts in this pull request

PR #946, opened 2026-06-25
feat(link): block links (FIP-263) with V19-gated type-scoped compaction

bond00729:
## Summary

Implements **FIP-263 (Block Links)**. A "block" is a `Link` with `type="block"` — it rides the existing Link CRDT, so there's no new message type, no new validation, and no separate quota (blocks share the follows quota).

The only runtime change is to stop link compaction from deleting across link types, so a `follow` compaction can't delete a user's `block` links. This mirrors Shuttle's type-scoped `getConflictCriteria`. The rest of the PR is test + conformance coverage.

## Why this needs a version gate

`merge_compact_state` deleted superseded links without checking link type. Type-scoping the deletion changes which messages get deleted, which changes the merkle trie **state root** in every shard header — so applying it ungated would fork a patched node on historical replay.

It's not a no-op on history, either: `unfollow`-typed links already exist in the data (a client bug — `unfollow` is 8 bytes and slips past `validate_link_type`, which has no allowlist). Under the old type-blind code a `follow` compaction would sweep them; under the fix it keeps them → divergent root.

So the fix is gated behind a new `EngineVersion::V19` / `ProtocolFeature::BlockLinks`, derived from the message's own timestamp. Pre-V19 replays byte-identical to today; V19+ is type-scoped. **Mainnet/testnet activation is a year-2100 placeholder** — it must be set to a real date after this reaches 100% of nodes.

## Changes

- **Compaction fix + gate** — `merge_compact_state` skips links whose type differs from the compact state's, gated by a `scope_compaction_by_type` flag. The engine derives the flag from the message's `EngineVersion`; the Store layer stays version-agnostic. (`merge_add`/`merge_remove` already scope by type structurally via a direct key lookup — only the prefix-scanning compaction path needed the filter.)
- **`EngineVersion::V19` + `ProtocolFeature::BlockLinks`** — placeholder activation, devnet from genesis.
- **Block coverage** (no production code — blocks already work) — store round-trips, follow/block independence, validation accepts `"block"`, and the cross-type compaction test now asserts blocks survive a follow compaction at V19.
- **Conformance** — `link_add_block` / `link_remove_block` / `link_compact_state_follow` golden vectors, plus a cross-node test that two V19 engines agree on the shard root after a block-add / follow-add / follow-compact sequence.

## Deploy ordering

Dormant until V19 activates, so safe to merge and deploy. Set the real V19 date (testnet before mainnet) only after this is at 100% of nodes.

---

> [!NOTE]
> **High Risk**
> Changes link compaction merge rules and state roots behind a version gate; incorrect gating or activation timing would fork nodes on replay, though pre-V19 behavior is explicitly preserved in tests.
> 
> **Overview**
> **FIP-263 block links** are supported as ordinary `Link` messages with `type="block"` (no new message type). Validation and merge for adds/removes stay on the existing Link CRDT; tests and conformance vectors pin that behavior.
> 
> The **behavioral fix** is **link compact state**: compaction used to sweep older links **without matching link type**, so a `follow` `LinkCompactState` could delete `block` (and other-type) links and change the shard merkle root. Compaction conflict logic now lives on **`LinkStore`’s `StoreDef`** and, when **`ProtocolFeature::BlockLinks`** is on (**`EngineVersion::V19`**), only deletes adds/removes whose type matches the compact state’s type.
> 
> **`Store::merge`** now takes a **`MergeContext`** (engine version from the message timestamp). Generic compact-state handling delegates to **`read_compact_state_details`**, **`is_compaction_conflict`**, and **`compact_state_add_conflict`** instead of hard-coded link parsing in `store.rs`. Shard/block engines thread `MergeContext` into all store merges.
> 
> **Versioning:** `V19` + `BlockLinks` on devnet from genesis; mainnet/testnet remain on V18 until scheduled. Pre-V19 replay keeps **type-blind** compaction; V19+ is **type-scoped**. Conformance corpus bumps to V19 and adds block link golden vectors plus cross-node compaction determinism tests.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8a4c9486c25de7d15627474e78be60b3ba8866d1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/version/version.rs
```diff
@@ -5,6 +5,11 @@ use strum_macros::EnumIter;
 
 const LATEST_PROTOCOL_VERSION: u32 = 13;
 
+// Placeholder V19 (Block Links) activation, far in the future, until the real rollout
+// date is set. Shared by the mainnet/testnet schedules and the schedule test so they
+// can't drift; split into real per-network dates at launch.
+const V19_PLACEHOLDER_ACTIVATION: u64 = 4102444800; // 2100-01-01 UTC
```
bond00729: @topocount not sure what a good cutover time is for the change

src/storage/store/account/store.rs
```diff
@@ -678,10 +681,11 @@ impl<T: StoreDef + Clone> Store<T> {
...
 
-        let (fid, compact_state_timestamp, target_fids) =
+        let (fid, compact_state_timestamp, target_fids, compact_state_type) =
             self.read_compact_state_details(message)?;
 
-        // Go over all the messages for this Fid, that are older than the compact state message and
+        // Go over all messages for this Fid older than the compact state (when type-scoped,
+        // only those of the compact state's own link type — see below) and
         // 1. Delete all remove messages
         // 2. Delete all add messages that are not in the target_fids list
```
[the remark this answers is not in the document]
bond00729: Fixed in 0bcdbf9 — reworded from "older than" to "not newer than" (the scan stops at `timestamp > compact_state_timestamp`, so equal-timestamp messages are still swept, i.e. `timestamp <=` the compact state). No behavior change.

src/storage/store/account/store.rs
```diff
@@ -564,6 +564,7 @@ impl<T: StoreDef + Clone> Store<T> {
         &self,
         message: &Message,
         txn: &mut RocksDbTransactionBatch,
+        scope_compaction_by_type: bool,
     ) -> Result<HubEvent, HubError> {
```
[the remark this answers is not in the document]
bond00729: Deferring this — reasonable suggestion, but I think a single threaded `bool` is the right call here rather than premature generalization.

`scope_compaction_by_type` is the **only** feature flag that crosses into `Store::merge`. The convention in this codebase is that version-gated behavior is resolved at the **engine** layer (where `EngineVersion` lives) and `Store` stays version-agnostic — e.g. `gasless_enabled` is a match guard in `merge_message` (`MessageType::KeyAdd if gasless_enabled`), and the other `is_enabled(...)` checks (EnsValidation, PrimaryAddresses, ReadDataFromShardZero, …) are all engine-side. This flag only had to cross the boundary because compaction scoping happens deep inside `merge_compact_state`'s prefix scan, where the engine can't intercept by message type.

So N=1 and likely to stay 1 — a `MergeOptions` struct for one self-documenting bool would be speculative. The natural trigger to introduce it is the day a *second* behavior flag needs to reach `Store::merge`; that change can convert this bool alongside it.

On the always-`false` threading: that is intentional, and I think the safer option. The alternative (deciding the scoping inside the link store) is the real footgun — it would silently diverge from the engine's version on replay. An explicit, greppable bool keeps the consensus-relevant decision visible and owned by the version-aware layer.

src/version/version.rs
```diff
@@ -651,7 +707,7 @@ mod version_test {
         let time = FarcasterTime::from_unix_seconds(1782147600);
         assert_eq!(
             EngineVersion::next_version_timestamp_for(&time, FarcasterNetwork::Mainnet),
-            None
+            Some(V19_PLACEHOLDER_ACTIVATION)
         );
```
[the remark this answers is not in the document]
bond00729: https://github.com/farcasterxyz/snapchain/pull/946#discussion_r3477945668

src/storage/store/account/store.rs
```diff
@@ -564,6 +564,7 @@ impl<T: StoreDef + Clone> Store<T> {
         &self,
         message: &Message,
         txn: &mut RocksDbTransactionBatch,
+        scope_compaction_by_type: bool,
```
topocount: I think there is a cleaner way to do this. Rather than threading this all the way through, store the logic inside the inner `store_def()` which is actually message type aware. You could store the compaction logic inside the inner store_def and call into it. We can add the protocol feature gate there as well, and we thread the engine version through in a `MergeContext { version }` struct, which allows us to implement future feature gates that are easier to read and cleaner to implement.

This way we aren't hoisting implementation details to the top and we aren't threading extraneous env info (like the `network`) down where it doesn't belong.

src/storage/store/account/store.rs
```diff
@@ -698,12 +703,17 @@ impl<T: StoreDef + Clone> Store<T> {
...
+                    message.data.as_ref().and_then(|data| data.body.as_ref())
+                {
+                    // When scope_compaction_by_type is set, only compact links of the compact
+                    // state's own type (so a "follow" compaction can't delete "block" links);
+                    // otherwise compaction is type-blind. The caller sets this from the engine
+                    // version (ProtocolFeature::BlockLinks).
+                    if !scope_compaction_by_type || link_body.r#type == compact_state_type {
+                        if self.store_def.is_remove_type(&message) {
+                            merge_conflicts.push(message);
+                        } else if self.store_def.is_add_type(&message) {
```
topocount: This feels leaky to me. Granted the original code wasn't clean to start but the message type unwrapping should not be accomodating link types explicitly. The link_store `store_def()` should determine what happens by parsing the message for us. Feel free to add an additional fn and member to the Store trait to facilitate this across all store types.

topocount (changes requested):

src/storage/store/account/store.rs
bond00729: Done in 31c6d8e. Swapped the bool for a `MergeContext { version }` that the engine builds once from the message's EngineVersion and passes to every store's merge. The feature gate now lives in the link store_def (`is_compaction_conflict` does the `ProtocolFeature::BlockLinks` check), so the engine just hands down the version and nothing link-specific gets hoisted up anymore.

Verified it's behavior preserving: the conformance state-root vectors and the cross-node determinism test are unchanged, so pre-V19 replay stays byte-identical.

src/storage/store/account/store.rs
bond00729: Agreed, moved it. 31c6d8e adds three StoreDef methods (`read_compact_state_details`, `is_compaction_conflict`, `compact_state_add_conflict`), all with default impls so only LinkStore overrides them. `Store<T>` doesn't unwrap `Body::Link*` anywhere now, the store_def owns all the parsing and the version gate.

One thing I decided: I used a concrete `CompactStateDetails` struct instead of an associated type on the trait. An associated type would force all 7 stores to declare it (associated type defaults are still unstable) even though only links do compaction, and it'd break the defaulted-method pattern. `Store<T>` only reads fid/timestamp off it and the link-specific fields flow straight back into the link hooks. Fine to switch to an associated type if you'd prefer, but concrete felt like less ceremony for a 1-of-7 case.

topocount (approved):
Looks great, added issue https://github.com/farcasterxyz/snapchain/issues/948 as a followup to some of this link type plumbing and to capture how we want to evolve this further.

Thanks @bond00729 !

PR #975, opened 2026-07-20
feat: admit and merge channel messages on shard 0

bond00729:
Makes the four channel message types real on shard 0 at V20: routed,
admitted under the channel authority model, and merged into per-channel
slot stores. Nothing fans out yet — channel state is invisible to data
shards and read APIs.

Stores (one row per logical slot; supersedes delete the incumbent even
cross-author, with trie/store/index/event agreement):
- ChannelUpdate: whole-replace per channel; unset membership/casting modes
  fold to the most restrictive value, so a cosmetic update can never
  silently reopen a channel.
- ChannelMember: one slot per (channel, fid) driven by a six-action state
  machine (member/moderator/removed/banned); removed and banned occupy
  their slots permanently, blocking resurrection.
- ChannelPin: one slot per channel (empty cast_hash unpins).
- ChannelModerate: one slot per (channel, cast_hash).
- Per-channel caps: 8192 member slots and 16384 moderation slots enforced at
  the slot-minting arms, and at most 10 live moderators enforced at admission.
- The 32-byte channel_id and 20-byte moderation cast_hash widths are
  enforced at validation AND at the store layer before any key is built —
  slot-key injectivity and cap integrity depend on them.

Authority (admission on shard 0, all reads through the caller's
transaction, no wall clock anywhere):
- The channel owner fid resolves registry owner_address → shard-0
  verification replica under a deterministic winner rule.
- A full (action × author-role × target-state) authority table implements
  the proposal's matrix with state-aware admission: moderators manage
  members but cannot demote or out-ban each other, UNBAN only applies to
  banned targets, never-seen fids cannot be minted by removal actions,
  self-leave is always allowed for members and moderators, self-join only on
  OPEN channels, and the owner cannot be banned. The table is enforced by an
  exhaustive cell-by-cell test matrix.
- Fail-closed parked semantics: while the owner address has no live
  shard-0 verification, every channel action is rejected with a distinct
  "channel is parked" reason — except self-leave, whose authority is
  self-contained. This keeps the owner-immunity rule total (no window
  where a moderator can ban an unresolvable owner) and means a channel
  recovers the moment its owner (re-)verifies.

Channel timestamps are deliberately inert (slots resolve by consensus
order), so there is no timestamp floor for channel types; the rationale is
documented at the validation arm.

---

> [!NOTE]
> **High Risk**
> Introduces consensus-critical shard-0 merge semantics, authority rules, and slot/counter invariants; mistakes could fork or corrupt channel state across nodes.
> 
> **Overview**
> **Channel messages become real on shard 0** when `ProtocolFeature::ChannelMessages` is on: stateless validation and the mempool stop rejecting the four channel bodies/types, routing always sends them to shard 0, and `BlockEngine` runs registry → owner-fid → role-based admission (including parked-channel freeze and moderator caps) before merging.
> 
> **New per-channel slot stores** (`ChannelUpdate`, `ChannelMember`, `ChannelPin`, `ChannelModerate`) use consensus-order supersede (not timestamp LWW), fixed-width keys, slot caps, and dedicated merge entry points; generic `Store::merge` paths are blocked via `requires_consensus_order_slot_merge`.
> 
> **Supporting changes:** RocksDB channel prefixes/postfixes, shared `select_verification_address_winner` and transactional verification-by-address reads for same-block authority, and RPC owner resolution updated to match. Gasless `KEY_ADD` scopes still explicitly deny channel types.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c0951b0898d2ab51a4eb2e9e6d11106a6a0d7a98. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/network/server.rs
+++ b/src/network/server.rs
@@ -130,7 +130,7 @@ fn signer_store_error_to_status(err: HubError) -> Status {
 fn resolve_channel_owner_fid(
     shard_stores: &HashMap<u32, Stores>,
     owner_address: &[u8],
-) -> Result<u64, Status> {
+) -> Result<Option<u64>, Status> {
     let mut authoritative_candidates = Vec::new();
 
     for stores in shard_stores.values() {
@@ -179,8 +179,7 @@ fn resolve_channel_owner_fid(
     Ok(
         crate::storage::store::account::select_verification_address_winner(
             authoritative_candidates,
-        )
-        .unwrap_or(0),
+        ),
     )
 }
 
@@ -2753,7 +2752,8 @@ impl HubService for MyHubService {
             .map_err(|err| Status::internal(format!("Store error: {:?}", err)))?
             .ok_or_else(|| Status::not_found("channel not registered"))?;
 
-        let fid = resolve_channel_owner_fid(&self.shard_stores, &channel_owner.owner_address)?;
+        let fid = resolve_channel_owner_fid(&self.shard_stores, &channel_owner.owner_address)?
+            .unwrap_or(0);
 
         Ok(Response::new(ChannelOwnerResponse {
             fid,
@@ -2789,7 +2789,8 @@ impl HubService for MyHubService {
         )?;
 
         if !channels.is_empty() {
-            let fid = resolve_channel_owner_fid(&self.shard_stores, &req.owner_address)?;
+            let fid =
+                resolve_channel_owner_fid(&self.shard_stores, &req.owner_address)?.unwrap_or(0);
             for channel in &mut channels {
                 channel.fid = fid;
             }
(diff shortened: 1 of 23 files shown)


src/network/server.rs
```diff
@@ -131,23 +130,14 @@ fn signer_store_error_to_status(err: HubError) -> Status {
...
-    // hash, so the higher `ts_hash` wins on timestamp and, at equal timestamps,
-    // on the message-hash bytes — this already resolves any tie between two
-    // distinct verifications. `Reverse(fid)` is only consulted on a full 24-byte
-    // `ts_hash` tie, which cannot occur for two distinct messages (the hash
-    // differs); it exists solely to keep the result total and deterministic.
-    // Strict `>` means an identical key never displaces the incumbent, so the
-    // winner is independent of shard/candidate iteration order (the
-    // `shard_stores` map iterates in an unspecified order).
-    let mut winner: Option<([u8; 24], Reverse<u64>)> = None;
+) -> Result<Option<u64>, Status> {
```
topocount, line 133: This should eventually be a set of owners

topocount (approved):

PR #996, opened 2026-07-31
feat(cli): add fc verification and devnet bootstrap commands

topocount:
Nothing in the repo could emit a verification message. `submit_message`, `perftest`, and the rest of `fc` produce CAST_ADDs only, which makes verification behavior impossible to exercise against a local node. This adds the two commands needed to do that.

Split out of the NEYN-12808 work (testing the M2 by-address migration on devnet) because it stands on its own.

## `fc verification add | remove`

```bash
export SIGNER_SECRET=0x<ed25519>    # the fid's registered signer
export WALLET_SECRET=0x<secp256k1>  # the address being verified

fc --node http://127.0.0.1:3483 --network devnet verification add --fid 200001

fc --node http://127.0.0.1:3483 --network devnet verification remove \
  --fid 200001 --address 0x<20-byte>
```

`VERIFICATION_ADD_ETH_ADDRESS` carries an EIP-712 claim binding `(fid, address, blockHash, network)`, signed by **the address being verified** — the node recovers the signer and requires it to equal the body address, so this is a bidirectional proof rather than an assertion by the fid. `WALLET_SECRET` (or `--wallet-secret`) can be omitted to derive from `MNEMONIC` at `--path` instead, matching how `key-add` gets its custody key.

Two constraints worth knowing:

- **EOA only** (`verification_type: 0`). Contract-signature verifications are intercepted in `submit_message_internal` and validated against a chain RPC outside the consensus loop, which a devnet has no endpoint for.
- The claim is built for **the network the message is finally submitted under**, not the factory's `Mainnet` placeholder. `retarget_network` rewrites the envelope but does not re-sign the claim, so a mismatch here produces a message the node rejects.

## `fc devnet bootstrap`

```bash
export SIGNER_SECRET=0x<ed25519>

fc devnet bootstrap --grpc-node http://127.0.0.1:3383 --auth dev:dev \
  --fid 200001 --count 60
```

The CLI's first `AdminService` command. A fresh devnet has no registered FIDs and nothing merges without one, so this submits the three on-chain events a usable FID needs — storage rent, ID-register, and a signer — and prints each FID's signer secret for piping into `verification add`.

ID-register is not incidental: DB migrations that enumerate FIDs do so from IdRegister events (`FIDIterator`), so a FID without one is invisible to them.

## On the duplicated EIP-712

The CLI mirrors the node's claim encoding by hand, as it already does for KeyAdd, to avoid depending on the snapchain crate at runtime. To keep that honest, `snapchain` is added as a **dev**-dependency — one-directional, since `snapchain` does not depend on this crate — and the tests run CLI-built messages through the node's real `validate_add_address` on all three networks, plus negative cases proving the claim is bound to its fid and its network.

A change to the domain, type list, or field encoding on either side now fails CI instead of silently emitting messages the node rejects.

## Testing

- `cargo test -p fc-cli` — 4 tests, all green.
- `make check-ci` clean on this branch standalone.
- Exercised end to end on a local devnet: bootstrapped 60 FIDs and submitted 60 verifications (20 addresses × 3 verifiers), against both a current node and a pre-#958 one, plus inside the Docker devnet.


topocount:
Thanks — addressed in c309fbc.

**Overflow / `--count 0`** — fixed, and it turned out to be worse than reported. Both guards were reachable only after `AdminServiceClient::connect`, so a bad invocation failed with a connection error instead of its own. Moved all argument validation (including signer-secret parsing) ahead of the connect:

```
$ fc devnet bootstrap --fid 1 --count 0
Error: "--count must be at least 1"
$ fc devnet bootstrap --fid 18446744073709551615 --count 5
Error: "--fid + --count overflows u64"
```

**EIP-712 domain comment** — rewritten. It was worse than ungrammatical, the field count was wrong: the declared `EIP712Domain` type lists four fields, while the domain populates only `name` and `version` and adds a `salt` that isn't declared. Now says that explicitly, and why the mismatch is deliberate.

**README table** — re-padded the whole column rather than just adding the space. `` `verification` `` is a character wider than the previous longest entry, so it was the alignment that broke, not only that one cell.

**`--auth` doc comment** — reworded. Small correction though: "bed" wasn't a typo for "binary", it was short for "test bed", i.e. a devnet created by `setup_local_testnet`. The phrasing was unclear either way, so it now reads "a devnet generated by `setup_local_testnet` uses whatever value that binary was given for `--admin-rpc-auth`".

`make check-ci` clean, 4 CLI tests green.

PR #1002, opened 2026-08-05
feat(cli): fc config pull — merge onchain-managed config from SnapchainConfigRegistry

matt-neynar:
Adds `fc config pull`, the consumer side of the SnapchainConfigRegistry (farcasterxyz/contracts#490/#491): fetch `configToml()` over a raw `eth_call`, structurally validate the three managed keys, and splice them into a local `config.toml` — replacing exactly `consensus.validator_sets`, `gossip.bootstrap_peers`, `gossip.direct_peers` and leaving everything else (`consensus.private_key` above all) untouched. Also adds `snapchain --check-config`, which loads and validates the config then exits 0/1, so deploy scripts can gate a restart on the merged file. NEYN-13024.

## Design notes

- **No new runtime deps except `toml`.** The fetch is a hand-rolled JSON-RPC POST over the existing `reqwest` + a `sol!` decode from the existing `alloy-sol-types`; the selector is pinned to the published ABI (`0x5a62bd75`) by a test.
- **Validation split.** `fc` mirrors `ValidatorSetConfig` locally (the crate's existing EIP-712 pattern) and validates structurally: hex + Ed25519 point decompression (the node unwraps both at startup), non-empty sets/shards, and `effective_at` monotonicity **per shard** — bounded against the most recent preceding entry sharing a shard, matching the contract; entries on disjoint shards are unordered. Full-config validation stays in the node via `--check-config`. Dev-dep parity tests pin the mirrors (validator sets against `validators.toml`, peer fields against the node's gossip config), and a round-trip test parses the mainnet history rendered through the registry's exact output grammar.
- **Bootstrap-peers opt-out (`--accept-local-bootstrap-peers-config`).** By default the pull replaces `gossip.bootstrap_peers` like the other managed keys, so nodes converge on the registry's canonical (public-routable) list. The flag opts a node out: a non-empty local list is operator-managed — e.g. VPC-internal addresses the registry's public list cannot carry — and survives the merge. Empty/whitespace local values adopt the registry list even with the flag; `consensus.validator_sets` and `gossip.direct_peers` are registry-authoritative in every mode. Deploy-side threading (env knob through apply script + watcher, so a flagged node's watcher compares like-for-like) lands with the C8/C9 wiring.
- **Network/registry selection.** Global `--network` became `Option<NetworkArg>` resolved per subcommand: message commands keep their testnet default; `config pull` defaults to mainnet. Registry resolution: `--registry` flag, else a baked-in per-network constant. Testnet is baked in (`0x00000000fc51aD6eb74EAE89ba4b01b1776fBA85` on Sepolia, verified against the live seeded deployment); mainnet deliberately stays `None` until the mainnet cutover — the same CREATE2 address will exist on L1, and baking it early would make the L1 contract deploy the act that starts mainnet pulls fleet-wide, ahead of the planned one-node dry-run gate. `--rpc-url` falls back to the config's `l1_rpc_url` on mainnet only — that URL exists for ENS and points at Ethereum mainnet on every node, so testnet requires an explicit Sepolia endpoint. An `eth_chainId` preflight (1 / 11155111; devnet unchecked) refuses reading a registry-shaped address on the wrong chain, and a declared `fc_network` that contradicts `--network` is a hard error.
- **Unattended-operation hardening** (this rewrites a file holding the consensus private key, from a container entrypoint): permission-preserving atomic replace (0600 temp, fsync file + dir, pid-unique name, cleanup on failure, symlinks resolved), HTTP timeouts + redirect ban + 4 MiB response cap, `--dry-run` redacts `private_key`, TOML parse errors never echo the source line, RPC URLs (which carry API keys) never reach logs or error strings, plaintext `http://` warns. Read-node configs are refused in write mode (the registry manages validators only).
- **ALPHA carve-out.** `config pull` is documented stable; the message-submission subcommands keep the ALPHA disclaimer, now scoped precisely.

## Verification

- 28 fc-cli tests (merge fixtures for both validator-set shapes the deploy paths write, empty-peer-is-a-value semantics, per-shard ordering in both directions, fail-closed unknown onchain keys, curve-point rejection, permission preservation) + node cfg tests.
- End-to-end against the real contract: compiled contracts `main` in a Foundry container, deployed + seeded on anvil (including a disjoint-shard lower-height entry), pulled from the host with `fc`, gated with `snapchain --check-config`. Merge, redaction, 0600 preservation, unseeded-registry refusal, and the wrong-chain guard all verified. Repro (adjust `$REG` to the printed deploy address):

  ```bash
  docker run -d --platform linux/amd64 --name snapreg-e2e -p 127.0.0.1:18545:8545 \
    --entrypoint sh ghcr.io/foundry-rs/foundry:latest -c "sleep infinity"
  docker exec snapreg-e2e sh -c "cd /tmp && git clone --depth 1 https://github.com/farcasterxyz/contracts.git \
    && cd contracts && git submodule update --init --recursive --depth 1 && forge build"
  docker exec -d snapreg-e2e sh -c "anvil --host 0.0.0.0 > /tmp/anvil.log 2>&1"
  # forge create SnapchainConfigRegistry, cast send appendValidatorSet x3 / setBootstrapPeers / setDirectPeers, then:
  fc --network devnet config pull --registry $REG --rpc-url http://127.0.0.1:18545 --config config.toml --dry-run
  snapchain --config-path config.toml --check-config
  ```

## Deferred / known limits

- The mainnet baked-in address is `None` until the mainnet cutover (see above); `--registry` is required on mainnet and devnet.
- `--check-config` is loader-deep: values that panic later in startup (e.g. a malformed `consensus.private_key`) pass it. Widening it is a deliberate scope decision, noted in a comment.
- The rewrite re-serializes through `toml::Table`, so comments in a hand-annotated config are dropped; preserving them would mean `toml_edit`.
- `rename(2)` cannot replace a single-file bind mount — deployments should mount the config's directory.
- Peer strings are validated as strings only; the node silently drops unparseable `PeerId`s, so content validation may be worth adding later.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -293,6 +293,14 @@ async fn main() -> Result<(), Box<dyn Error>> {
         }
     };
 
+    // Deploy scripts run `snapchain --check-config` between rewriting config.toml
+    // (fc config pull) and restarting the node, so a bad config fails here — before
+    // the running node is stopped — rather than as a crash loop after.
+    if app_config.check_config {
+        println!("config OK");
+        return Ok(());
+    }
+
     let env_filter = EnvFilter::try_from_default_env().unwrap_or_else(|_| EnvFilter::new("info"));
     match app_config.log_format.as_str() {
         "text" => tracing_subscriber::fmt().with_env_filter(env_filter).init(),
--- a/src/main.rs
+++ b/src/main.rs
@@ -293,14 +293,6 @@ async fn main() -> Result<(), Box<dyn Error>> {
         }
     };
 
-    // Deploy scripts run `snapchain --check-config` between rewriting config.toml
-    // (fc config pull) and restarting the node, so a bad config fails here — before
-    // the running node is stopped — rather than as a crash loop after.
-    if app_config.check_config {
-        println!("config OK");
-        return Ok(());
-    }
-
     let env_filter = EnvFilter::try_from_default_env().unwrap_or_else(|_| EnvFilter::new("info"));
     match app_config.log_format.as_str() {
         "text" => tracing_subscriber::fmt().with_env_filter(env_filter).init(),
@@ -311,6 +303,17 @@ async fn main() -> Result<(), Box<dyn Error>> {
         _ => {
             return Err(format!("Invalid log format: {}", app_config.log_format).into());
         }
+    }
+
+    // Deploy scripts run `snapchain --check-config` between rewriting config.toml
+    // (fc config pull) and restarting the node, so a bad config fails here — before
+    // the running node is stopped — rather than as a crash loop after. Placed after
+    // the log_format check so everything main() itself validates is covered; note
+    // the gate is loader-deep only — values that panic later in startup (e.g. a
+    // malformed consensus.private_key) are not exercised here.
+    if app_config.check_config {
+        println!("config OK");
+        return Ok(());
     }
 
     if app_config.clear_db {
--- a/src/main.rs
+++ b/src/main.rs
@@ -305,11 +305,31 @@ async fn main() -> Result<(), Box<dyn Error>> {
         }
     }
 
+    if app_config.statsd.prefix == "" {
+        // TODO: consider removing this check
+        return Err("statsd prefix must be specified in config".into());
+    }
+
+    let (statsd_host, statsd_port) = match app_config.statsd.addr.split_once(':') {
+        Some((host, port)) => {
+            if host.is_empty() || port.is_empty() {
+                return Err("statsd address must be in the format host:port".into());
+            }
+            Ok((host.to_string(), port.parse::<u16>()?))
+        }
+        None => Err(format!(
+            "invalid statsd address: {}",
+            app_config.statsd.addr
+        )),
+    }?;
+
     // Deploy scripts run `snapchain --check-config` between rewriting config.toml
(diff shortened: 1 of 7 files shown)


matt-neynar:
Doc note: `docs/snapchain-config-registry.md` §10 in the contracts repo still says both registries deploy to L1 mainnet — stale relative to the testnet-on-Sepolia decision in the C7 ticket. Code here follows the ticket (Sepolia, chain id 11155111).

src/main.rs
```diff
@@ -305,23 +305,6 @@ async fn main() -> Result<(), Box<dyn Error>> {
         }
     }
 
-    if app_config.clear_db {
```
topocount, line 308: why is this getting deleted

topocount (dismissed):
approved with one question about the db clearing flag logic getting deleted. Seems out of scope

src/main.rs
matt-neynar: Not deleted — moved. 7a4ccb0 hoisted the statsd validations above the new `--check-config` gate so the gate covers everything main() itself validates, and `--clear-db` stayed behind the gate (a validation run must never touch the database). The diff renders the reorder as a delete + re-add ~30 lines down; the block itself is unchanged.

topocount (approved):

PR #1005, opened 2026-08-06
C8: apply-onchain-config.sh plus image and compose wiring

matt-neynar:
Part of NEYN-13025 (FIP-263 stack). **Stacked on #1002 (C7)** — base is the C7 branch so only C8's commits show; will retarget to main and rebase once C7 merges.

## What

- **`scripts/apply-onchain-config.sh`** — startup helper the compose entrypoints call between the config heredoc and `exec $0 $@`: runs `fc config pull`, validates the merge with `snapchain --check-config`, caches the result as last-known-good on a volume, and on pull/validation failure boots from that cache (loudly) rather than letting an L1 RPC blip take a validator down — an outage during a rolling restart could otherwise cost quorum. Hard-fails (crash-loop via `restart: always`) only when there is neither a fresh pull nor a cache. No-ops on read nodes (`read_node` is the only mode flag; registry manages validator config only).
- **Stub-based test harness** (`scripts/apply-onchain-config-test.sh`, 29 checks: skip paths, success + cache perms, fc argument construction, fallback, refuse-to-boot). Both files shellcheck-clean.
- **Dockerfile** — copy the script into `/app` (with a `.dockerignore` negation; `/scripts` is otherwise excluded). Also fixes the stale default `CMD ["./snapchain", "--id", "1"]` (`--id` isn't a flag; the bare default exited 1).
- **Compose** — wire the script into `docker-compose.mainnet.yml` / `docker-compose.testnet.yml` entrypoints with env threaded via compose-time interpolation (empty defaults, so plain `docker compose up` is unchanged) and a dedicated `.onchain-config` volume for the cache.

The Neynar-validator wiring (`.stack/` in farcasterxyz/snapchain-deployer) is a separate PR in that repo.

## Deviation from the ticket, for review

The ticket frames the escape-hatch flag as opt-out (script on by default). This PR makes it **opt-in** (`ONCHAIN_CONFIG_ENABLED`, default off): until C5/NEYN-13022 bakes registry addresses into `fc`, every pull fails — and with no cache yet written, the ticket's own hard-fail rule would crash-loop every validator on first deploy. Opt-in makes shipping C8 inert; C10's cutover flips it on, and leaving it unset stays the operator escape hatch for pinned configs.

## Known tradeoff

The last-known-good cache is the **full merged config** (re-extracting just the three registry keys in bash would reimplement C7's merge). On a fallback boot, env-derived values (secrets, RPC URLs) revert to their last-pull state until the RPC recovers; the script logs this explicitly.

## Behavior change for public testnet operators

This PR adds `restart: always` to `docker-compose.testnet.yml` (mainnet and the deployer already had it). It is required for the C9 watcher's TERM-based restart cycle and also retries a refused boot — but it changes standalone behavior for anyone relying on a testnet node staying down after an exit, independent of whether the onchain-config feature is enabled.

## Testing

- `./scripts/apply-onchain-config-test.sh` — all 29 checks pass (macOS bash; stubs for `fc`/`snapchain`)
- `shellcheck` clean on both scripts
- `docker compose config -q` passes for both compose files; `.dockerignore` negation verified with a throwaway image build
- Not yet run: the single-node anvil e2e rig from NEYN-13027 (needs C5's registry deploy or a local registry contract)

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -7,3 +7,4 @@
 /Makefile
 README.md
 /scripts
+!/scripts/apply-onchain-config.sh
--- a/.dockerignore
+++ b/.dockerignore
@@ -6,5 +6,7 @@
 /Dockerfile
 /Makefile
 README.md
-/scripts
+# Exclude the directory's children individually (not the directory itself) so
+# the negation below works under every .dockerignore matcher implementation.
+/scripts/*
 !/scripts/apply-onchain-config.sh
--- a/Dockerfile
+++ b/Dockerfile
@@ -74,6 +74,9 @@ COPY --from=builder \
   /usr/src/app/target/release/perftest \
   /usr/src/app/target/release/fc \
   /app/
+COPY scripts/apply-onchain-config.sh /app/
 
 ENV RUSTFLAGS="-Awarnings"
-CMD ["./snapchain", "--id", "1"]
+# Compose files override both entrypoint and command; this default only covers
+# a bare `docker run` (the previous `--id 1` was not a valid flag and exited 1).
+CMD ["./snapchain", "--config-path", "config.toml"]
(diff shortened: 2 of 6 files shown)


topocount (dismissed):
Please just modify the shell logs to emit in roughly the same json shape as snapchain itself by default. Otherwise looks good

matt-neynar:
Done in 6285b6f — log() now emits the node's tracing_subscriber JSON shape ({"timestamp","level","fields":{"message"},"target"}) with tracing's level names. Also applied to onchain-config-watch.sh on the C9 branch (a182909), which had the same bare-text logging.

topocount (approved):

